PR #8, opened 2015-07-22
New Dashboard layout with 3 columns (date completion, rdt, info)

ivanarrizabalaga:


app/src/main/java/org/eyeseetea/malariacare/layout/adapters/dashboard/AAssessmentAdapter.java
```diff
@@ -62,73 +64,26 @@ public View getView(int position, View convertView, ViewGroup parent) {
...
-        ((TextCard) rowView.findViewById(R.id.score)).setText(getStatus(survey));
+        //INFO
+        TextCard info = (TextCard) rowView.findViewById(R.id.info);
+        info.setText(survey.getValuesToString());
 
+        rowView.setBackgroundResource(LayoutUtils.calculateBackgrounds(position));
         return rowView;
     }
 
-    /**
```
adrianq: @ivanarrizabalaga We are going to have sth like a progress bar (2 of 7) maybe it is a good idea to not delete this method and just change the implementation. However you know the code better than me so does it make sense?

PR #26, opened 2015-08-03
Questions font size, Khmer i10n, DB updated

ivanarrizabalaga:


adrianq:
@ivanarrizabalaga there are a bunch of .idea and app.iml files that change in this PR. Were there changed on purpose? I guess no but just to understand if there is a reason for.

PR #39, opened 2015-08-17
Multidex enabled

ifoche:


ifoche:
I've reverted a previous commit in development as apparently I broke development by introducing crashlytics. So in this PR, the modifications are:
- Crashlytics introduction
- MultiDex library introduction

There is a bug inside when accessing a survey that is related to the DB so please do not merge this PR yet, until this bug is solved

ifoche:
Bug solved, apparently it is now safe to merge it

Issue #69, opened 2015-09-22
Update strings to English


adrianq:
Apparently, just the numbers but please @rhardjono check all the strings

Issue #68, opened 2015-09-22
Text from previous question in textfield


adrianq:
The client said:
I completed a survey, sent it and then tried to review my responses. Interestingly, to the right is a screenshot of the final question (Q7) of one of my sent surveys. As you can see, the answer is ‘in the plantation?’, which must have generated from my response to the previous question (I certainly didnt type it). I saw this same error during our call yesterday for the ‘Age’ question (Q3) but had difficulty describing it. This must be the same bug. Glad I caught a screenshot of it.

Issue #63, opened 2015-09-21
Text field improvements


adrianq:
- [x] Keyboard is still displayed after completing the age field and phone field
- [x] Numeric keypad automatically display when moving to these questions
- [x] Set the input field background color to light grey

ifoche:
[These are the lines](https://github.com/EyeSeeTea/malariapp/blob/v1.1/app/src/main/java/org/eyeseetea/malariacare/SurveyActivity.java#L519-521) where this is done in malariapp

adrianq:
Many thanks @ifoche!

rhardjono:
do you mean keyboard has to hide after press enter button...?

ifoche:
In malariapp, keyboard is hidden when focus is lost. I don't know if you
want that in malariapp but I just wanted to clarify that. For me, that
makes sense

rhardjono:
@ifoche thanks!

Issue #60, opened 2015-09-21
Cancel should work as the negative button


adrianq:
Store an X ( this value needs to be confirmed by the client) and end the survey

rhardjono:
have to show the confirmation dialog too?

adrianq:
Yes, same behaviour as the No option

adrianq:
@rhardjono can you make sure we show this option as X (code) in the dashboard? The name should be 'Not tested'

rhardjono:
Ok, i'll make this change.

PR #70, opened 2015-09-24
Review/bugfix/keypad

rhardjono:
#63 #60 #68 #69 #47

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/layout/adapters/survey/DynamicTabAdapter.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/layout/adapters/survey/DynamicTabAdapter.java
@@ -42,6 +42,8 @@ import android.view.LayoutInflater;
 import android.view.MotionEvent;
 import android.view.View;
 import android.view.ViewGroup;
+import android.view.WindowManager;
+import android.view.inputmethod.InputMethodManager;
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/eyeseetea/malariacare/layout/adapters/survey/DynamicTabAdapter.java
```diff
@@ -398,15 +423,32 @@ public void onClick(View v) {
...
+
+        //Take focus to this view
+        editText.requestFocus();
+        editText.postDelayed(new Runnable() {
+           @Override
+           public void run() {
+               //Show keypad
+               showKeyboard(ctx, editText);
+           }
+       }, 300); //use 300 to make it run when coming back from lock screen
```
adrianq: @rhardjono can you explain me a little bit more what this 300 means?

rhardjono: This value is the time you wait until the keypad is shown.
A value lower than 300 make it not working properly (for example when
coming back from lock screen.) but it can be adjusted if needed...

2015-09-25 12:57 GMT+02:00 Adrian Quintana <notifications@github.com>:

Issue #75, opened 2015-10-07
Implement number picker for age question (Default: 0)


adrianq:
See [this link](http://developer.android.com/reference/android/widget/NumberPicker.html) for further information

PR #78, opened 2015-10-08
Implement number picker for age question (Default: 0)

idelcano:
Change the edit text by number picker.
Change the function initPositiveIntValue to work with numberPicker. 
Suggest refactor to initAgeValue.
Tested in Api 16,21 and 23.


adrianq:
@idelcano Note there are two different things: [the card which contains the issue](https://github.com/EyeSeeTea/pictureapp/issues/75) and the card which is automatically generated once you create the pull request (PR), this one. The first one is assigned to you and you are the one in charge of adding any comment, changing the status, etc. The second one, as I said, is automatically created when you create the PR together with the comments you add to the PR and is my responsibility. The card will be moved, again automatically, to the 'To Test' column and at this moment I will be in charge of adding comments or merging the PR and close the card

btw, thanks for the PR! :)

PR #85, opened 2015-10-10
Button remove all sent results

idelcano:
I forgot commit and push the changes of the file  with the Shared Preferents.


idelcano:
The file is the xml with the differents options. The aplication crash without it.

PR #88, opened 2015-10-13
Auto complete editext org unit

idelcano:
Tested in API 16 && API 21


idelcano:
Fixed. i forgot commit the "git rm" before push.
It´s ready
I tested in real device Android 5.0.1

app/src/main/java/org/eyeseetea/malariacare/EyeSeeTeaApplication.java
```diff
@@ -45,6 +46,7 @@ public void onCreate() {
         Fabric.with(this, new Crashlytics());
         LocationMemory.getInstance().init(getApplicationContext());
         // Not previously populated tables
+        Session.setUIDPROGRAM(getResources().getString(R.string.UID_PROGRAM));
```
adrianq: @idelcano Why are you storing this uid in the session? Maybe it is better to read it [like this](https://github.com/EyeSeeTea/malariapp/blob/master/app/src/main/java/org/eyeseetea/malariacare/network/PushClient.java#L175) as we discussed.

idelcano:
It´s true, It´s fixed, sorry.

PR #89, opened 2015-10-15
Extract imei phone number

idelcano:
@adrianq look first the comment of the string.xml.

It was test in API 16 API 23 and in real device in v5.0.1


adrianq:
Ok, now I understand the problem. I would like to have a look at it a little bit more in detail before merging this PR. I would give it a try tonight and come back to you  so we can find the best solution

idelcano:
@adrianq i merged somefiles at manual way for first time, in think is correct. I was revising all and i think is all ok for merge.

PR #109, opened 2015-10-21
Fixed delete survey

idelcano:
Fixed some bugs and issues:
-The user saves only the first time , the rest is updated 
-Pull the phoneMetadata in one dataelement 
-Update License for every class
-Review swype (delete) on dashboard.
-All deleted survey.class were failing because the model had an invalid value
-Cancel PushAlarm if PushClient is not valid. The application was still working underground sending requests (AlarmPushReceiver.class ).
-The surveys negative  were interpreted as false, i change it, and now they are sent. [link](https://github.com/EyeSeeTea/pictureapp/commit/420081831f5d4da21ea6bd2d284514ef33e371a1)

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/database/AppDatabase.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/database/AppDatabase.java
@@ -1,3 +1,23 @@
+/*
+ * Copyright (c) 2015.
+ *
+ * This file is part of QIS Survelliance App.
+ *
+ *  QIS Survelliance App is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  QIS Survelliance App is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with Foobar.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
 package org.eyeseetea.malariacare.database;
 
 import com.raizlabs.android.dbflow.annotation.Database;
--- a/app/src/main/java/org/eyeseetea/malariacare/database/AppDatabase.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/database/AppDatabase.java
@@ -1,4 +1,21 @@
-
+/*
+ * Copyright (c) 2015.
+ *
+ * This file is part of QIS Survelliance App.
+ *
+ *  QIS Survelliance App is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  QIS Survelliance App is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with QIS Survelliance App.  If not, see <http://www.gnu.org/licenses/>.
+ */
 
 package org.eyeseetea.malariacare.database;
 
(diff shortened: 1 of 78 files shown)


app/src/main/java/org/eyeseetea/malariacare/network/PushClient.java
```diff
@@ -113,8 +111,8 @@
     //When PushClient is sending the event, the activity is null, becouse PushClient is called in a InstanceService without activity.
     //I can´t access to the activity or context for get the r.string values.
     private static String TAG_IMEI="RuNZUhiAmlv";
```
adrianq: @idelcano it is a minor thing but maybe we can use a better name than TAG_IMEI as now it is going to be the uid for three fields. So that in the future we know what we are storing here.

app/src/main/java/org/eyeseetea/malariacare/database/AppDatabase.java
```diff
@@ -1,3 +1,23 @@
...
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  QIS Survelliance App is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with Foobar.  If not, see <http://www.gnu.org/licenses/>.
```
adrianq: @idelcano I think this Foobar should be removed with the name of the app as well. I have just realised it is wrong for as well in malariapp

idelcano:
@adrianq this pull request is incomplete.

I tested on the development server, and I found not optimized calls, as well as some bug fixes. (I fixed it)
But I'm getting this exception when i call dev.psi-mis.org:

exception: java.net.ProtocolException: Too many follow-up requests: 21
call: https://dev.psi-mis.org/api/programs/IrppF3qERB7.json?fields=organisationUnits

In malariacare.psi.org works all right.
cal:https://malariacare.psi.org/api/programs/IrppF3qERB7.json?fields=organisationUnits

I have to find a solution, but until the evening I can not work today.
I also realized that the changing of the server URL is not updated in the application without restart.

adrianq:
@idelcano understood. I leave it open until you get a chance to fix it. I have made a few comments but actually I haven't gone through all the code. It looks good what it is there at the moment by the way. :)

Regarding the exception, can you paste it here for me together with any other information you can find relevant? I think it is a problem of the server so I would like to check it with @josemp10. I am not sure but I think it is not your fault but something we have to configure on the server.

The other problem you mention (not updating the url until app is restarted) is certainly a bug so we need to fix it before merging this PR. 

Thanks for the work! I think it would be good to schedule a meeting for Friday so we can review everything plus a couple of comments the client will let me know in the following hours. Let's talk by chat about this.

josemp10:
Looking at the error, it seems that @idelcano user is not reaching the
server, or there is some issue regarding the internet connection of dev

@idelcano, could you please send to me the json file you are pushing to the
server?

Thanks

idelcano:
@ josemp10 @adrianq 
i'm sorry, the server side works ok. The problem was pictureapp and malariaapp BasicAuthentication was different. (picture app without user, malariaapp with user).

Anyway, i have this response when i try to pull a survey:
{"httpStatus": "OK", "HttpStatusCode": 200, "status": "OK", "message": "Import was successful." "Response": {"responseType": "ImportSummaries","imported":0,"updated":0,"ignored":0,"importSummaries":[{"responseType":"ImportSummary","status":"ERROR","description":"**Current user does not Have This Program permission to access** "," importCount ": {" imported ": 0," updated ": 0," ignored ": 0," deleted ": 0}}]}}
I have no permissions to send surveys (user:idelcano).

another question for @adrianq .
Pictureapp not have user login.

I'm not know what I have to do. 
 
some options:
1. Continue accesing to the  test server( the user/password autentication is not required, the development server not work).

2. Overwrite the BasicAuthentication method, and sets my user, or other user, as default user. And access with user/password authentication (test server not work).

3. Try connect without authentication, if it fails, try to connect with authentication.

4. Only check if it works in development server, and comment the code for future release.

The default user in the aplication is user/user.

I think for now I just have to check it out on the development server.
In that case my user does not have sufficient permissions to send surveys. 

I will work in the bug of the change of URL now.
Maybe you prefeer talk about it in the next meeting.

josemp10:
@idelcano, I see. It is a problem of permissions, but what facility are you
pushing to? The program is only activated on testing facilites 'KH Outlet
1-1', 'KH Outlet 1-2', 'KH Outlet 1-3'....

josemp10:
Can you try again @idelcano? I have changed your permissions in dev... Let
me know if it works...

idelcano:
Great @josemp10 . it works now.

I was pushing to KH Outlet 1-2.
I can only use KH Outlet 1-2 because the rest do not have the code element .
[Organisations Units in our program](https://dev.psi-mis.org/api/programs/IrppF3qERB7.json?fields=organisationUnits)

idelcano:
@adrianq it´s ready for merge.

PR #126, opened 2015-10-26
Force remove some uses permissions.

idelcano:


idelcano:
I try in android api 16(emu) and api 21(real device) and the aplication works  without that uses-permission.

adrianq:
@idelcano Looks good to me but before merging into development I think it would be good to discuss the implications of this change. I will keep it open until we have a hangout.

PR #127, opened 2015-10-26
Change ban organization logic

idelcano:
The application counts how many surveys Have Been created in the same hour for each of the surveys between 0 and 1 hours.

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/network/PushClient.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/network/PushClient.java
@@ -122,7 +122,7 @@ public class PushClient {
 
     private static String TAG_PHONEMETADA="RuNZUhiAmlv";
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/eyeseetea/malariacare/network/PushClient.java
```diff
@@ -122,7 +122,7 @@
 
     private static String TAG_PHONEMETADA="RuNZUhiAmlv";
 
-    private static int DHIS_LIMIT_SENT_SURVEYS_IN_ONE_HOUR=30;
+    private static int DHIS_LIMIT_SENT_SURVEYS_IN_ONE_HOUR=10;
```
adrianq: @idelcano We should set this back to 30, right?

idelcano: Right. I´m sorry.

PR #134, opened 2015-10-28
About us text

idelcano:
@adrianq i copy the structure of malariapp 

I found a bug, if you press the link mailito the aplication crash. I think It needs be called in a Intent


idelcano:
I added 
        alertDialog.setCancelable(false);
to make the dialogs model dialogs.

Issue #147, opened 2016-01-07
Ask for login/password when setting a new OU or server url


adrianq:
We are going to use the SDK login system. Every time the user wants to change the OU or server url, user and password will be asked. However, push continue as current:  it won't be asked during the push process, and will use the current hard coded username/ password.

adrianq:
Related to #148

adrianq:
@ivanarrizabalaga As we discussed on our previous meeting there were a couple of unclear points. Let me try to shed some light on them:
-  When the user clicks either on the OU or on the url field, the login window should be displayed.
- The login window should display three fields: user, login and server url. As it is in the current sdk login window. The server url should be the current one, if any, although it should be editable.
- It would be good to remember the last user name (not the password) so the second he wants to change anything the username is filled by default together with the server url. Again this field should be always editable at any time.
- If the user provides a correct user/pw the field should become editable. Once he exits the preferences interface, these fields are not editable anymore and user/pw will be asked again if he comes back to the preference interface and tries to change OU or url.
- In order to retrieve the list of available OU we should use the hidden/internal user/pw. We only use the user credentials for allowing the user to change the fields but for the rest of the communication with the server (push, OU retrieving, etc) we have to use the internal/hidden one.

I think we have covered everything and this makes much more sense but, of course, if you think something is not clear just give me a shout

Issue #156, opened 2016-01-25
Move API actions from PushClient into its class


ivanarrizabalaga:
PushClient should be an assistant to push data to the server via API.

Little by little it has turned into a bag of non related api calls (not only Push).
There must be 2 classes.

 - PushClient: for pushing 2.20
 - APIClient: for direct callsto API (serverVersion, checkOrgUnit, ...)

Issue #153, opened 2016-01-13
Verify number of imported elements on push


adrianq:

adrianq:
https://github.com/EyeSeeTea/malariapp/pull/679/files

Issue #151, opened 2016-01-07
SDK integration


adrianq:
Integrate SDK into the application and use SDK pull/push feature. Application should be compatible with DHIS SDK 2.21. Tests will be performed targeting 2.20 and 2.21 DHIS2 servers (2.22 probably by the end of the release). For servers 2.20, most of the current mechanisms to retrieve and push information will be used. However for 2.21 and 2.22 we have to use the SDK as much as possible in order to communication to the server,

- [x] Identify server version on init /server url change/orgunit. API call: https://hnqis-staging.psi-mis.org/api/system/info
- [x] Implement mechanism in order to allow switching between current behaviour for 2.20 server: Push and Org Unit & Url selection
- [x] Migrate and adapt visitors.
- [x] For 2.21 and above implement:
   - [x] When changing OU and url, download metadata (available OU) and once selected data for the last three months.
   - [x] Push data

Note there are a few API calls in the [PushClient class](https://github.com/EyeSeeTea/pictureapp/blob/1e7948206be314c2f8fb4ee5d877834fb4f3ed38/app/src/main/java/org/eyeseetea/malariacare/network/PushClient.java) that should stay there after the SDK integration. These calls allow to ban a OU when too many surveys has been pushed, checked if a OU is open, etc.

ivanarrizabalaga:
@adrianq The default user 'KHMCS' is not ready in *https://hnqis-staging.psi-mis.org*, I guess this is going to be required to test 2.21 sdk.

adrianq:
@josemp10 Can you please create [this user](https://github.com/EyeSeeTea/pictureapp/blob/1e7948206be314c2f8fb4ee5d877834fb4f3ed38/app/src/main/java/org/eyeseetea/malariacare/network/PushClient.java#L97-L99) in staging so that we can start testing the sdk integration? @ivanarrizabalaga In the meanwhile, can you use your user? Do you have permissions for the OU/server?

adrianq:
@ivanarrizabalaga  A few details to keep in mind:
- Staging will be updated to 2.22 by the mid-end of this week. We will probably also have two dedicated servers with 2.22.
- For  now keep working with sdk 2.21 branch targeting 2.21 staging server. However, once the server is migrated to 2.22 we have to change  to 2.22 sdk branch. I am aware it can be a sort of a pain but I think most of the work will be reused as difference between the two branches (and server versions) are minimum.
[...]

PR #159, opened 2016-01-31
Sdk 221

ivanarrizabalaga:
Closes:
 #147, #148, #151 , #153 , #156 , #158

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/SettingsActivity.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/SettingsActivity.java
@@ -99,6 +99,10 @@ public class SettingsActivity extends PreferenceActivity implements SharedPrefer
         setupSimplePreferencesScreen();
     }
 
+    void setAutoCompleteEditTextPreference(AutoCompleteEditTextPreference autoCompleteEditTextPreference){
+        this.autoCompleteEditTextPreference=autoCompleteEditTextPreference;
+    }
+
     /**
      * Shows the simplified settings UI if the device configuration if the
      * device configuration dictates that a simplified, single-pane UI should be
@@ -382,7 +386,7 @@ public class SettingsActivity extends PreferenceActivity implements SharedPrefer
     @TargetApi(Build.VERSION_CODES.HONEYCOMB)
     public static class GeneralPreferenceFragment extends PreferenceFragment {
 
-        private AutoCompleteEditTextPreference autoCompleteEditTextPreference;
+//        private AutoCompleteEditTextPreference autoCompleteEditTextPreference;
         @Override
         public void onCreate(Bundle savedInstanceState) {
             super.onCreate(savedInstanceState);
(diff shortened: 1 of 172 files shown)


adrianq:
@ivanarrizabalaga when trying to edit the preferences, the login window shows up properly however url textfield is not editable. Can you please make it editable as discussed?

ivanarrizabalaga:
Sure, my fault. The worst is that somehow I did it on purpose ugggg

adrianq:
np, try to keep it what you have at the moment (comment it out or sth like this). As you know user can change its mind while testing it.

app/src/main/java/org/eyeseetea/malariacare/database/iomodules/dhis/importer/PullController.java
```diff
@@ -0,0 +1,311 @@
...
+     * Notifies that the pull is over
+     */
+    private void postFinish() {
+        //Fixme maybe it is not the best place to reload the logged user.(Without reload the user after pull, the user had diferent id and application crash).
+        User user = User.getLoggedUser();
+        Session.setUser(user);
+        try {
+            Dhis2Application.getEventBus().post(new SyncProgressStatus());
+        } catch (Exception e) {
+            e.printStackTrace();
```
adrianq: @ivanarrizabalaga to print the stack trace is fine in general for this app, however I think it would make sense to show a kind of warning dialog during the pull process. I think this is what we are doing for most of the exceptions in the pull, right? What do you think?

app/src/main/java/org/eyeseetea/malariacare/SettingsActivity.java
```diff
@@ -260,11 +440,11 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
     @TargetApi(Build.VERSION_CODES.HONEYCOMB)
     public static class GeneralPreferenceFragment extends PreferenceFragment {
 
-        private AutoCompleteEditTextPreference autoCompleteEditTextPreference;
         @Override
         public void onCreate(Bundle savedInstanceState) {
             super.onCreate(savedInstanceState);
             addPreferencesFromResource(R.xml.pref_general);
+            addPreferencesFromResource(R.xml.pref_general);
```
adrianq: @ivanarrizabalaga why adding this twice?

app/src/main/java/org/eyeseetea/malariacare/SettingsActivity.java
```diff
@@ -63,11 +75,38 @@
...
     protected void onCreate(Bundle savedInstanceState) {
+        Dhis2Application.bus.register(this);
         super.onCreate(savedInstanceState);
     }
 
+    @Override
+    public void onStop(){
+        try {
+            Dhis2Application.bus.unregister(this);
+        }catch(Exception e){}
```
adrianq: @ivanarrizabalaga Maybe we can show a warning dialog?

app/src/main/java/org/eyeseetea/malariacare/database/iomodules/dhis/importer/PullController.java
ivanarrizabalaga: @adrianq, I would remove the try/catch.

If you get this far the pull process is completed, this is just a mechanism to communicate (via bus) the progress activity to move on and show a proper dialog. An exception at this point should never happen.

If this happens there is something bigger not related to the pull itself and the app will break sooner or later.

app/src/main/java/org/eyeseetea/malariacare/SettingsActivity.java
ivanarrizabalaga: @adrianq this is just an awful copy&pase error, this code was running for just for tablets. Now it is fixed.

app/src/main/java/org/eyeseetea/malariacare/SettingsActivity.java
ivanarrizabalaga: @adrianq the try/catch is not required, an exception should never be raised here.

A warning dialog might be a little bit confusing since the app will get back to the dashboard ( you're leaving the settingsactivity) and the potential (never expected) error is not related to you or anything you can do about it.

PR #234, opened 2016-06-09
Bug s4 from master fix bugs

idelcano:
@ifoche i create this branch with some minus bugs resolved in S5.

-Fix the undefined RTD value.
-Fix the crash if read a value without option(like Time capture).
-Fix  bug pulling events to old( return/continue).


ifoche:
@idelcano I'm not going to include this as 4.1 is going to be abandoned and introducing these fixes to master branch directly will cause later conflicts for the next PR to master. If S5 weren't yet in testing process I would.

PR #249, opened 2016-06-10
Bug fix remove old values

### LINKED ISSUE #245: Review the values when a parent question changes
All the values are saved and sent at this moment.

idelcano:
Closes #245


ifoche:
@idelcano I think you need to refactor this PR a little bit. A few considerations that comes quick to my mind:

- Please do not use ReadWriteDB class, it was thought to disappear as soon as we will find the time to refactor its code. The functions that you put there can go inside the pojos. 

- You have an empty if condition. What's the purpose of that?
```
if (value!=null){

}
```

- Can you rename showValues() function to printValues()?

idelcano:
Yes, so sorry, i think its much better now

PR #276, opened 2016-06-15
Remove backPressed on survey review backbutton

idelcano:
when you close a survey in sentfragment and the isBackPressed is set as true, the next new survey completed  will be removed and not send.
It is solved in this PR.


ifoche:
@idelcano do you finally consider this as definitive?

idelcano:
@ifoche Yes, it works,  i decided add "isBackPressed=false" in initSurvey to be sure 100%.
Anyway, I tested it  many times.

PR #284, opened 2016-06-15
Send the Date and time in the event Timecapture and Time Sent

### LINKED ISSUE #278: Time sent and Time capture are reporting the date but not the time

idelcano:
Closes #278
@ifoche It works for my in the first try. maybe i missing something?
For example the event: https://dev.psi-mis.org/api/events/h81VkWjSVDm
In the web the date is shown with american format.


ifoche:
I don't know, they might have solved it in this last month. Thanks !

PR #312, opened 2016-07-05
Added lao/thai font and move khmerOS to cambodia

### LINKED ISSUE #311: Find & test lao and thai fonts

idelcano:
closes #311 
Added font: https://fontlibrary.org/es/font/boon-v1-1 SIL Open Font License, Version 1.1. http://scripts.sil.org/cms/scripts/page.php?item_id=OFL_web


ifoche:
@idelcano so finally, khmerOS.ttf was not enough to represent all the languages? why are you introducing boon.ttf here? I see that it looks like an open source font library, though.

PR #354, opened 2016-08-03
Fix reason and threatment screens (text size, question type..)

idelcano:
@adrianq this PR is about screen cosmetic(text size, text position, screen type..) I will explain you in the meeting.


adrianq:
@idelcano Merging this one. You know better than me what is going on here :)

PR #381, opened 2016-08-12
Fix bug, get survey from query not from session on reloadAnswered

### LINKED ISSUE #380: reloadAnswered survey from Session crash on restart.
Is necessary get the survey from db not from session.

idelcano:
Closes #380


app/src/main/java/org/eyeseetea/malariacare/database/model/Survey.java
```diff
@@ -865,6 +866,18 @@ public void removeChildrenValuesFromQuestionRecursively(Question question) {
             }
         }
     }
+    /**
```
adrianq: @idelcano can you move this method to Survey.java?

PR #382, opened 2016-08-12
Fix pull change server before acept eula

### LINKED ISSUE #374: Bug on first pull(Server version not valid)
-  [x] Reload url in first pull
- [x] Add dialog to show unexpected server error on server version check
- [x ] Change default server
The server check is failing in www.psi-mis.org with the Lao user.

This is the server error:

getServerVersion: java.net.ProtocolException: Too many follow-up requests: 21

We will change the default server to dev.psi-mis in lao variant.

idelcano:
Closes #374


app/src/main/java/org/eyeseetea/malariacare/LoginActivity.java
```diff
@@ -134,11 +134,12 @@ private Intent propagateExtraAndResult(Intent intent){
         }
 
         if(getIntent().getBooleanExtra(SettingsActivity.SETTINGS_EULA_ACCEPTED, false)){
-            Log.i(TAG, "propagateExtraAndResult -> EULA accepted");
+            Log.i(TAG, "propagateExtraAndResult -> EULA accepted, Server overwrite from "+PreferencesState.getInstance().getDhisURL() +" to "+getUserIntroducedServer());
+            PreferencesState.getInstance().setDhisURL(getUserIntroducedServer());
```
adrianq: @idelcano the server introduced by the user in the login screen should not replace the server in the settings? Is this line changing this behaviour?

PR #404, opened 2016-08-29
Feature change psk to psi strings

### LINKED ISSUE #403: change PSK to PSI string in laos

idelcano:
Closes #403
@ifoche  i fixed the PSI number to(in lao and cambodia)... Only the english string had the correct phone.


idelcano:
Closes #403

PR #408, opened 2016-08-29
Set as sent only if the survey send was finish(like in malariapp).

### LINKED ISSUE #405: inconsistent monitor surveys
Checks if a survey can be placed in monitoring and removed from it, in the survey status workflow.

idelcano:
Closes #405
I removed unused method, and a survey sets as sent only when was sent like in malariaapp

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/database/model/Survey.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/database/model/Survey.java
@@ -747,13 +747,6 @@ public class Survey extends BaseModel  implements VisitableToSDK {
         }
     }
 
-    public void updateSurveyState(){
-        //Change status and save mainScore
-        setStatus(Constants.SURVEY_SENT);
-        save();
-        saveMainScore();
-    }
-
     public static void removeInProgress() {
         List<Survey> inProgressSurvey= getAllUncompletedSurveys();
         for(int i=inProgressSurvey.size()-1;i>=0;i--){
--- a/app/src/main/java/org/eyeseetea/malariacare/network/PushClient.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/network/PushClient.java
@@ -199,9 +199,6 @@ public class PushClient {
     }
 
     public void updateSurveyState(){
-        //Change status
-        this.survey.setStatus(Constants.SURVEY_SENT);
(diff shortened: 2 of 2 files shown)


ifoche:
@idelcano this code only removes the survey state change...but when is the survey being set to SENT? I've tested the code and the surveys that I send never get reflected in the monitor module...did you forget to push some code?

idelcano:
the survey sets as sent when the important summary returns a correct upload without conflicts and with out conexión fails

idelcano:
@ifoche here: https://github.com/EyeSeeTea/pictureapp/blob/95a7f5753a6126aebc7f51e0ec8240ca34576c53/app/src/main/java/org/eyeseetea/malariacare/database/iomodules/dhis/exporter/ConvertToSDKVisitor.java#L285

ifoche:
can you test it then and tell me if looking at the monitor module, if you get the numbers updated with the new survey you've just sent?

ifoche:
Maybe a monitor module reload would be necessary after setting the survey as sent?

idelcano:
But the survey only sets as sent when the push controller send it, It sets
a survey as completed-》wait for  push controller when push controller
finish if the survey was pushed it change the survey state to send and
monitor only shows the sent surveys.  is it not correct ?

idelcano:
@ifoche I checked it again and it works as expected.

- When you complete a survey it is set as completed.
- The pushcontroller try to push all the completed surveys.
- When a survey is pushed, the converter changes the survey state to sent.
- The monitoring tab shows the sent surveys. (Maybe do you want show the completed surveys too?).

In this commit i removed  action to change the survey state to sent before push. We remove it in malariaapp some time ago. And this resolves this problem (If the push fails some surveys can change his status from sent to completed and disappear from monitoring).

ifoche:
@idelcano, I've checked it again and it looks like the only problem is the automatic reload of the monitor. I think when you change the state of the survey we need to reload the monitor tab, otherwise unless you change the tab and back to monitor you wont see that change. Can you please do that?

ifoche:
As we have just said by chat, I'm accepting this, not allowing to reload in order to avoid getting the dropdown and all the tab reloaded, as if the dropdown was configured to 6 days and with the reload it changes to 6 months, the user may not notice and get confused.

PR #414, opened 2016-08-31
Feature review fragment changes

### LINKED ISSUE #411: Modify reviewFragment
- [x] Click in one question in the reviewfragment should move the screen to that question.
- [x] Move the finish dialog from phone question to reviewFragment.

idelcano:
Closes #411


ifoche:
@idelcano fill in a survey with a Pv/PvPf positive case, and no phone number. Then on the review screen instead of send just click on "review". This is taking me to a warning instead of taking me to the first question...can you check it and solve it?

idelcano:
yes, i will do it now. i think is necessary skip the reminder( and not move
it to the first question, move it to the last question, the Phone question
in this case).

idelcano:
@ifoche it is solved

PR #423, opened 2016-09-06
Feature option padding

### LINKED ISSUE #409: Add properties for image in layouts and finish with text overlapping bug
The Reason  question and the Treatment  question doesn't show the text properly in some resolutions. The text is mixed with image.

Suggestion: add padding proportional according to the screen

idelcano:
Closes #409
Tested in:
480x854 tvdpi 4,5" [LG leon](http://www.lg.com/es/telefonos-moviles/lg-Leon-4G-LTE-H340N)
1080x1920 420dpi 5" tablet similar to the test team samsung  tablet
720x1080 hdpi  5"
240x320 ldpi 2,7"


ifoche:
@idelcano closing this PR as it finally was not accepted as a good solution. We will reuse this ideas in case it finally becomes a need for the client.

PR #426, opened 2016-09-12
Monitoring Question ids and options ids names refactor

### LINKED ISSUE #422: Review monitor id names and separe ids for each flavour

idelcano:
closes #422


ifoche:
@idelcano just testing this branch and it looks like something is broken. Concretely, if you try to send a survey, if it's not tested it apparently reach the server but is not the case for the tested. And it doesn't looks like a server problem because I don't even see the send procedure happening (so there's no conflict in the origin of this problem). Can you please test it asap? Thanks

ifoche:
As we just said by videoconference I'm merging this into development in order to let you solve that problem in a different branch as it probably doesn't have nothing to do with the IDs

PR #428, opened 2016-09-13
fix translations

ifoche:


ifoche:
@idelcano translations integrated into development. You can proceed with MigrationTable whenever you want

PR #432, opened 2016-09-23
Change images

### LINKED ISSUE #431: Include new treatment image with embedded text

idelcano:
Closes #431


ifoche:
@idelcano did you check the review screen? Now seems that the treatment option is not being reflected there. And it's quite normal. If I'm not bad we were using the text of the question to represent it (as for some options the text is sent in english while it's shown in lao) and now the text is empty...What do you think? 

I think in 2 different possibilities:
1 - maybe it would be better to keep the text but introduce a new variable to determine if the question is type: (IMAGE_TYPE || IMAGE_TEXT_TYPE || TEXT_TYPE) and this way we would represent the image, image + text or just the text depending on the type of question...

2 - Another quicker solution (and probably the best due to the lack of time we have to provide a new release) is to just check in the review screen if the question text is empty, and represent the code in that case. 

Could you add the second solution unless you think 1st one is very quick to implement? (moving this to in progress in the meanwhile)

ifoche:
@idelcano. Table migration is not properly working. It's adding the questions text but not modifying the alignments to 3;3, so text is both shown in the image and as a floating text if you come from the previous version. Can you please take a look to this whenever you can?

Thanks

idelcano:
@ifoche  i forgot increase the migration number. I fixed and tested it again. It's done.

ifoche:
@idelcano thanks!

PR #436, opened 2016-09-28
Fix repeated survey

idelcano:


ifoche:
@idelcano this was not yet completed, right? If I'm not bad from what we talked yesterday, certain tasks must still be address in this branch. Otherwise please let me know if I can finish merging this branch

PR #460, opened 2016-10-11
Feature move login to start

### LINKED ISSUE #443: Move login to app start

Strabagkantza:
Closes #443 Move from master to development

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/LoginActivity.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/LoginActivity.java
@@ -19,25 +19,41 @@
 
 package org.eyeseetea.malariacare;
 
+import android.app.AlertDialog;
 import android.app.LoaderManager.LoaderCallbacks;
+import android.content.Context;
+import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.Loader;
 import android.content.SharedPreferences;
 import android.database.Cursor;
 import android.os.Bundle;
-import android.preference.Preference;
+
 import android.preference.PreferenceManager;
+import android.text.Html;
+import android.text.SpannableString;
+import android.text.method.LinkMovementMethod;
+import android.text.util.Linkify;
 import android.util.Log;
 import android.view.MenuItem;
 import android.view.View;
 import android.widget.EditText;
+import android.widget.TextView;
+import android.widget.Toast;
 
 import com.squareup.otto.Subscribe;
 
+import org.eyeseetea.malariacare.database.iomodules.dhis.importer.PullController;
+import org.eyeseetea.malariacare.database.model.User;
+import org.eyeseetea.malariacare.database.utils.PopulateDB;
 import org.eyeseetea.malariacare.database.utils.PreferencesState;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/eyeseetea/malariacare/LoginActivity.java
```diff
@@ -61,24 +77,41 @@
...
 
-        //Username, Password blanks to force real login
-        EditText usernameEditText = (EditText) findViewById(R.id.username);
-        usernameEditText.setText("");
-        EditText passwordEditText = (EditText) findViewById(R.id.password);
-        passwordEditText.setText("");
+       SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(this);
+       EditText passwordEditText = (EditText) findViewById(R.id.password);
+       EditText usernameEditText = (EditText) findViewById(R.id.username);
+       if (sharedPreferences.getString("","").isEmpty()) {
```
ifoche: @Strabagkantza why this if? I mean, you're not asking for as string to the preferences, but putting "" where the key should be. That's always going to answer "" right? Is there any situation in which this would give a different answer?

app/src/main/java/org/eyeseetea/malariacare/SettingsActivity.java
```diff
@@ -916,6 +916,7 @@ void launchLoginOnEulaAccepted(){
     }
 
     Intent prepareIntent(){
+        ProgressActivity.PULL_CANCEL =true;
```
ifoche: @Strabagkantza correct me if I'm wrong but you're here avoiding the pull after the login. So now there's no pull. Can you explain this a little bit more? Maybe put a comment detailing your intentions?

app/src/main/res/values/donottranslate.xml
```diff
@@ -12,8 +12,10 @@
     <item name="customize_fonts" type="string" translatable="false">customize_fonts</item>
     <item name="font_sizes" type="string" translatable="false">font_sizes</item>
     <item name="dhis_url" type="string" translatable="false">dhis_url</item>
+    <item name="dhis_user" type="string" translatable="false">dhis_user</item>
+    <item name="dhis_password" type="string" translatable="false">dhis_password</item>
     <item name="org_unit" type="string" translatable="false">org_unit</item>
-
+    <item name="pull_metadata" type="string" translatable="false">pull_metadata</item>
```
ifoche: @Strabagkantza You added this string to the donottranslate file but you never use it. Why? BTW, remember you're here adding it to the main branch, and not to the myanmar variant...do you need it to be here?

ifoche: @Strabagkantza as we have just talked, we would also need to:
- [ ] Add pull to login (only data)
- [ ] Remove Login from settings
- [ ] Add Logout to settings
- [ ] Add EULA acceptance to Login

ifoche:
@Strabagkantza closing this PR because it doesn't contain all the commits we're needing to solve the issue, as you put the new commits in nother branch (see #466)

PR #507, opened 2016-11-14
Feature refactor settings activity

### LINKED ISSUE #500: Refactor SettingsActivity

xurxodev:
closes #500


xurxodev:
@ifoche pending changes are pushed in this branch. For laos and cambodia SettingsActivity should pull data when change server url or orgunit.

ifoche:
@xurxodev this code produces a NullPointerException for me when dealing with Session.java user (if you click on server setting, first time just after you install the app from zero, when you still doesn't have any user on DB). It's trying to make a Logout, and as part of the logout try to remove the user from Session...causing a crash. Does it happen for you or is it a result of something I forgot to include when merging development into the branch?

idelcano:
@xurxodev  @ifoche 
I found this problem too. 
i think the problem is not in the xurxodev code, the problem is we forget to convert the UserAccount to our User.

The User is not converted in pictureapp at this moment and never is in our database.

we need add the user visitor to userAccount:

UserAccountExtended  userAccountExtended = new UserAccountExtended(MetaDataController.getUserAccount());
        userAccountExtended.accept(converter);

here: 
https://github.com/EyeSeeTea/pictureapp/blob/development/app/src/main/java/org/eyeseetea/malariacare/database/iomodules/dhis/importer/PullController.java#L335

ifoche:
@idelcano more or less, I think you're right saying we were not setting that user because we didn't need it for Laos and Cambodia, so it's not we forgot to add the visitor, but we didn't need it. But obviously if the new Settings uses the logout function, needs to create the user in order to work. I'll let @xurxodev answer to listen his opinion. My guess is that on Laos and Cambodia apps there was no explicit logout and so nobody was trying to delete the user (hence, no crash), but now Myanmar introduced this logout concept on dev change inherited from malariapp, and then we need to manage this situation translating the user too.

xurxodev:
@ifoche @idelcano I think this bug is in this code in sesión.java 
```
public static synchronized void setUser(User user) {
        Log.i(TAG, "Creating user in session. " + user.toString());
        Session.user = user;
    }
```
Log.i not verify null for user, It is fixed in move login to app start branch. I did not think the source of the bug was on this branch.

ifoche:
@xurxodev In my Android monitor what's crashing is concretely [this line](https://github.com/EyeSeeTea/pictureapp/blob/8acc50b8525990daf642475b8bf867de456f8d14/app/src/main/java/org/eyeseetea/malariacare/database/utils/Session.java#L143-L143) on logout() function. That Log.i() may also be a problem but definitively we have also a problem in logout(). I'll wait for your confirmation tomorrow when you will have the time to test it. Thanks in advance

PR #510, opened 2016-11-16
Feature move login to app start

### LINKED ISSUE #443: Move login to app start

xurxodev:
closes #443


xurxodev:
@ifoche ok as they are not the definitive languages I thought it was not necessary

PR #514, opened 2016-11-21
modify the importSummary survey status strategy

### LINKED ISSUE #513: Duplicates on intermittent connection
The user is able to create duplicates by connecting and disconnecting WiFi while surveys are being sent.

idelcano:
Closes #513 
The solution is setting the survey status as quarantine(to be checked) when the survey has a importSummary.
If the importSummary is invalid( F.E. network problems), the survey should be quarantine.


idelcano:
@ifoche  i test it again, i think its correct

ifoche:
@idelcano I've tested 2 or 3 times, and I've managed to reach the following situation:
- I sent 7 surveys, with intermittent connectivity. 
- I deliberatively cut the connexion while sending, while checking the quarantine and on every possible situation of the whole send process.  
The 7 surveys have been successfully uploaded to the server, no duplicates (that's great), but app side I have 6 surveys on sent tab and one survey on unsent tab. Looking at the messages it looks really strange because it's like that survey is on quarantine, but when the app check it's in the server it doesn't take it out from the quarantine. Please, take a look to this screenshot:
quarantine

Can you take care of this? I think we're very close to the final solution

idelcano:
@ifoche I fixed the bug and i added a reloadDashboard on launchQuarantineChecker to refresh  the unsentFragment  if some surveys was move from quarentine to sent status

PR #581, opened 2016-12-09
Feature add progressbar

### LINKED ISSUE #558: Incorporate Progress bar to Login
This progress bar would disappear very quickly from Laos/Cambodia variants but would stay on Myanmar all the time the populateDB is working

idelcano:
Closes #558


idelcano:
@ifoche 
Is impossible include all the methods used in EyeSeeTeaApplication and in Dhis2Application inside of the SpashScreenActivity (used to show a progressbar wheel).
I added a icon before the SpashScreenActivity, and i think it looks great, (or  at least better).

PR #621, opened 2016-12-13
Reminder and counter image align to left

### LINKED ISSUE #613: Align invalid question image to the left

xurxodev:
closes #613


xurxodev:
@ifoche this feature was for counter question or for rdt question? I fixed counter question and reminder question because is same layout but I remember we had talked about aligning left in the rdt question but I do not see that issue

PR #636, opened 2016-12-14
Remove send/review popup and instead of that, put a cancel button on …

### LINKED ISSUE #604: Remove send/review popup and instead of that, put a cancel button on the review screen

xurxodev:
…the review screen. Hide cancel button for surveys with max counters reached
closes #604


xurxodev:
@ifoche Changes are ready, user can review survey with invalid rdt but not increase counters again

xurxodev:
@ifoche changes are ready, now in review mode not show counter screen but save and move to next question then change selected option is enabled.

PR #655, opened 2016-12-15
Bug scroll jump top

### LINKED ISSUE #634: Scroll always take you to the top of the screen

idelcano:
Closes #634


ifoche:
I accepted this PR without testing it given we're a bit stressed to finish other features. Thanks @idelcano for your explanation about why was that 30px offset happening on the scroll event

PR #709, opened 2017-01-23
Feature new sdk login logout

### LINKED ISSUE #708: Login and Logout with new sdk in Clean Architecture way

xurxodev:
closes #708


app/src/main/java/org/eyeseetea/malariacare/data/database/iomodules/dhis/importer/ConvertFromSDKVisitor.java
```diff
@@ -40,7 +40,8 @@
 import java.util.List;
 import java.util.Map;
 
-public class ConvertFromSDKVisitor implements IConvertFromSDKVisitor {
+public class ConvertFromSDKVisitor implements
```
idelcano: @xurxodev  Only to be consistent: 
it is not so important but can you remove "org.eyeseetea.malariacare.data.database.iomodules.dhis.importer." and "org.eyeseetea.malariacare.data.database.model.XXX" and use only one import in the entire class?
I think we are using the imports instead of the entire path except when the SDK model have the same name than our model and they are used in same class, but in the new sdk we dont have this problem.

app/src/main/java/org/eyeseetea/malariacare/data/database/model/Option.java
```diff
@@ -57,15 +57,15 @@
     /**
      * Reference to parent answer (loaded lazily)
      */
-    Answer answer;
+    org.eyeseetea.malariacare.data.database.model.Answer answer;
```
idelcano: @xurxodev in this class too

app/src/main/java/org/eyeseetea/malariacare/domain/exception/NetworkException.java
```diff
@@ -0,0 +1,8 @@
+package org.eyeseetea.malariacare.domain.exception;
+
+
+public class NetworkException extends Exception {
+    public NetworkException() {
+        super("Credentials not valid");
```
idelcano: @xurxodev  can you change "Credentials not valid" by "Network error" or similar? thanks

xurxodev:
@idelcano I have apply changes. Can you review again please?

PR #929, opened 2017-03-13
Feature make compile cambodia and laos

### LINKED ISSUE #829: Cambodia and lao not compiling because no found strings that exists for myanmar, it related to split treatment matrix csv

idelcano:
closes #829


app/src/cambodia/res/values/donottranslate.xml
```diff
@@ -32,9 +32,35 @@
...
     <string name="patientResidenceVillageOtherCode" translatable="false"></string>
-
     <string name="control_data_element_phone_metadata" translatable="false">RuNZUhiAmlv</string>
     <string name="control_data_element_datetime_capture" translatable="false">qWMb2UM2ikL</string>
     <string name="control_data_element_datetime_sent" translatable="false">aBahytzj2u0</string>
-    <string name="category_cc" translatable="false"></string>
+    <string name="error_no_treatment" translatable="false"> </string>
+
+    <string name="nonexistent" translatable="false">nonexistent</string>
+    <string name="dynamicTreatmentHideQuestionUID" translatable="false">@string/nonexistent</string>
```
xurxodev: @idelcano if there is no treatment matrix for lao and cambodia, do not you think it would be better not to contaminate them with treatment matrix strings and code? I think it would be better if all that has to do with treatment matrix exists within strategy only for myanmar.

idelcano:
@xurxodev  it's done

idelcano:
sorry @xurxodev it is fixed

PR #937, opened 2017-03-14
Feature refactor reload survey answered ratio

### LINKED ISSUE #724: Refactor reloadSurveyAnsweredRatio() to take Stock Program into account

manuelplazaspalacio:
Close #724


app/src/main/java/org/eyeseetea/malariacare/data/database/model/Survey.java
```diff
@@ -989,6 +1009,85 @@ public SurveyAnsweredRatio reloadSurveyAnsweredRatio() {
         return surveyAnsweredRatio;
     }
 
+    private SurveyAnsweredRatio reloadStockSurveyAnsweredRatio() {
```
xurxodev: @manuelplazaspalacio DBFlow survey model is a class with a lot of responsibilities. All the new code that we do is better to separate it and little by little leave here only the columns of the table. I think it would be possible to add the logic of calculating the ratio of questions answered in domain and not in the data layer. So the new code is easier to split into methods to make it cleaner.

xurxodev (changes requested):
@manuelplazaspalacio I have request you any changes

manuelplazaspalacio:
@xurxodev  Refactor reload answer ratio to ReloadAnsweredRatioUseCase

app/src/main/java/org/eyeseetea/malariacare/data/database/model/Survey.java
```diff
@@ -311,6 +314,21 @@ public void setEventUid(String eventuid) {
...
 
     /**
+     * Ratio of completion is cached into answeredQuestionRatio in order to speed up loading
+     */
+    public SurveyAnsweredRatio getAnsweredQuestionRatio() {
+        if (answeredQuestionRatio == null) {
+            answeredQuestionRatio = SurveyAnsweredRatioCache.get(this.getId_survey());
+            if (answeredQuestionRatio == null) {
+                AReloadSurveyAnsweredRatioUseCase
+                        reloadSurveyUseCase = new ReloadSurveyAnsweredRatioUseCase(this);
```
xurxodev: @manuelplazaspalacio  an UseCase can not be invoked from data layer, Use case should be invoked only from presentation layer. On this way database model has more of one resposability. A possible option is from where we invoke getAnsweredQuestionRatio actually inkove ReloadSurveyAnsweredRatioUseCase

app/src/main/java/org/eyeseetea/malariacare/data/database/model/Survey.java
```diff
@@ -1006,10 +1008,12 @@ public void updateSurveyStatus() {
             return;
         }
 
-        SurveyAnsweredRatio answeredRatio = this.reloadSurveyAnsweredRatio();
+        AReloadSurveyAnsweredRatioUseCase reloadSurveyUseCase =
```
xurxodev: @manuelplazaspalacio an UseCase can not be invoked from data layer, Use case should be invoked only from presentation layer. On this way updateSurveyStatus has more of one resposability. A possible option is from where we invoke updateSurveyStatus first we calculate SurveyAnsweredRatio and pass it as parameter to updateSurveyStatus

app/src/main/java/org/eyeseetea/malariacare/data/database/model/Survey.java
```diff
@@ -311,6 +314,21 @@ public void setEventUid(String eventuid) {
     }
 
     /**
+     * Ratio of completion is cached into answeredQuestionRatio in order to speed up loading
+     */
+    public SurveyAnsweredRatio getAnsweredQuestionRatio() {
```
xurxodev: @manuelplazaspalacio SurveyAnsweredRatio is a domain concept not a data concept than cache management is better that be realized by new use case and not by database model

xurxodev (changes requested):
@manuelplazaspalacio I have request you some changes, can you review it please?

manuelplazaspalacio:
Close #724 @xurxodev

xurxodev:
closed because it has a lot of time, conflicts and it does not make sense

PR #938, opened 2017-03-14
Adding test organisation if demo user logged

### LINKED ISSUE #853: test_organisation should only be populate in demo mode but not in login mode

manuelplazaspalacio:
Close #853


app/src/main/java/org/eyeseetea/malariacare/data/database/datasources/AuthenticationLocalDataSource.java
```diff
@@ -83,6 +82,13 @@ private void clearCredentials() {
         PreferencesState.getInstance().reloadPreferences();
     }
 
+    private void addTestOrganisation() {
```
xurxodev: @manuelplazaspalacio Adding a dummy organization does not belong to the login process, it belongs to the pull process when the credentials are demo. You can move it to the pull controller where dummy organizations are added

PR #981, opened 2017-03-23
Feature make lao cambodia monitoring shown unsent too

### LINKED ISSUE #926: Introduce Announcement on kh and laos

### LINKED ISSUE #928: Make laos & cambodia "Run in demo mode"
Last-hour change. They don't want to take into account the unsent surveys, but to introduce a "Run in demo mode" option in the menu. Once pressed, the data would be erased and the mockup process from Myanmar would be used, to simulate the send process.

Please take a look at what we can still save from this PR, and determine whether we can reuse this PR or if it's better to start a new branch for this last-hour change.

idelcano:
Closes #928


ifoche:
Last-hour change. They don't want to take into account the unsent surveys, but to introduce a "Run in demo mode" option in the menu. Once pressed, the data would be erased and the mockup process from Myanmar would be used, to simulate the send process.

I'm reassigning this issue to @xurxodev. Please take a look at what we can still save from this PR, and determine whether we can reuse this PR or if it's better to start a new branch for this last-hour change.

idelcano:
The client changed this feature and this PR is not necessary

PR #1007, opened 2017-03-28
Feature run in demo mode cambodia and laos

### LINKED ISSUE #928: Make laos & cambodia "Run in demo mode"

idelcano:
Closes #928


app/src/cambodia/java/org/eyeseetea/malariacare/strategies/BaseActivityStrategy.java
```diff
@@ -19,16 +36,74 @@ public void onStop() {
...
+                runDemoMode();
+                break;
+            default:
+                return false;
+        }
+        return true;
+    }
+
+    public void runDemoMode() {
+        PopulateDB.wipeSurveys();
```
xurxodev: @idelcano according to the clean architecture, presentation layer should not call directly to the data layer to clean the database, better invoke LogoutUseCase which already clean the database and org units from preferences

PR #1018, opened 2017-03-29
Fix eReferrals variant

### LINKED ISSUE #1017: Fix eReferrals variant

manuelplazaspalacio:
Close #1017


app/src/ereferrals/res/values/donottranslate.xml
```diff
@@ -69,5 +69,8 @@
     <string name="category_option_group_set_uid" translatable="false">yVH67wmhryq</string>
     <string name="test_organisation_name">test_organisation</string>
     <string name="test_organisation_uid">9fV2JoHmO94</string>
+    <string name="GPS_PROVIDER_DEFAULT" translatable="false">default</string>
+    <string name="GPS_LATITUDE_DEFAULT" translatable="false">16.010790</string>
+    <string name="GPS_LONGITUDE_DEFAULT" translatable="false">96.852446</string>
```
ifoche: These are the coordinates for Myanmar. Maybe we can move this to some point in the lake Victoria, given that this app is mainly for Tanzania, Uganda and Mozambique

app/src/ereferrals/java/org/eyeseetea/malariacare/domain/usecase/CompletionSurveyUseCase.java
```diff
@@ -1,91 +1,7 @@
...
-    }
-
-    private Survey getSurveyWithStatusAndAnsweredRatio(long idSurvey) {
-        org.eyeseetea.malariacare.data.database.model.Survey surveyDB =
-                org.eyeseetea.malariacare.data
-                        .database.model.Survey.findById(idSurvey);
-        Survey survey = new Survey(idSurvey);
-        survey.setSurveyAnsweredRatio(surveyDB.reloadSurveyAnsweredRatio());
-        surveyDB.updateSurveyStatus();
-        survey.setStatus(surveyDB.getStatus());
```
ifoche: Are you sure that removing this part won't cause your survey to never pass to complete state? Probably the name of the function is badly chosen because a function named getSomething() should never do a permanent change in the DB, but anyway I think we need to keep it and rename the function. Unless you tell me that you moved elsewhere...but if that's the case I don't see that change in this PR.

manuelplazaspalacio: This is done within the DashboardActivityStrategy in the sendSurvey () and completeSurvey () methods. In Myanamar it works differently for the stock survey.

manuelplazaspalacio:
@ifoche changes requested done.

PR #1074, opened 2017-04-07
Feature refactor ban org unit code

### LINKED ISSUE #977: Refactor ban OrgUnit code
- [ ] abstract ban org unit code inside  push use case. Use Case should have not dependencies to external layers
- [ ] simplify isSurveysOverLimit method, if previously we order surveys list, only is necessary compare first and last survey

xurxodev:
closes #977


xurxodev:
@ifoche I have rename to OverLimitSurveysDomainService

xurxodev:
@ifoche sorry commit text is not good

PR #1100, opened 2017-04-24
Feature convert import sumary to entity and make push like hnqis

### LINKED ISSUE #1099: Add import summary from hnqis and push refactor

idelcano:
Closes #1099


xurxodev:
@idelcano I have changed base branch to v1.1_la

app/src/main/java/org/eyeseetea/malariacare/domain/entity/PushReport.java
```diff
@@ -0,0 +1,85 @@
+package org.eyeseetea.malariacare.domain.entity;
```
xurxodev: @idelcano I think it would be convenient to group the concepts related to pushreport within the same package, for example Push, PushResponse or PushSummary or another similar name

app/src/main/java/org/eyeseetea/malariacare/domain/entity/SurveyConflict.java
```diff
@@ -0,0 +1,28 @@
+package org.eyeseetea.malariacare.domain.entity;
+
+public class SurveyConflict {
```
xurxodev: @idelcano really uid from this class not is survey id then I think a name as PushConflict or simply Conflict is more appropriate

app/src/main/java/org/eyeseetea/malariacare/domain/exception/PushReportException.java
```diff
@@ -0,0 +1,20 @@
+package org.eyeseetea.malariacare.domain.exception;
```
xurxodev: @idelcano I think it would be convenient to group the exceptions related to pushreport within the same package, for example Push or another similar name

app/src/main/java/org/eyeseetea/malariacare/data/remote/PushDhisSDKDataSource.java
```diff
@@ -91,6 +102,44 @@ public void call(Throwable throwable) {
                 });
     }
 
+
+    private Map<String, PushReport> convertImportSummaryToPushReport(
```
xurxodev: @idelcano Can you extract this method to mapper class and create a unit tests for verify map or conversion is valid?

PR #1122, opened 2017-04-26
added language selection list translation if some language is selected

### LINKED ISSUE #1107: There's no translation for font size strings
When we moved to EyeSeeTea-sdk we broke translations for those strings. We need to fix it. Take a look to [the picture](https://cloud.githubusercontent.com/assets/17549893/25211799/39caef6a-25b1-11e7-9b38-65267b56c148.png)

idelcano:
Closes #1107

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/SettingsActivity.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/SettingsActivity.java
@@ -33,13 +33,11 @@ import android.preference.PreferenceActivity;
 import android.preference.PreferenceFragment;
 import android.preference.PreferenceManager;
 import android.util.DisplayMetrics;
-import android.util.Log;
 
 import org.eyeseetea.malariacare.data.database.utils.PreferencesState;
 import org.eyeseetea.malariacare.strategies.SettingsActivityStrategy;
 import org.eyeseetea.malariacare.utils.Utils;
 import org.eyeseetea.malariacare.views.AutoCompleteEditTextPreference;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/eyeseetea/malariacare/SettingsActivity.java
```diff
@@ -156,6 +154,35 @@ private static void bindPreferenceSummaryToValue(Preference preference) {
...
+        CharSequence[] newEntries = new CharSequence[5];
+
+
+        newEntries[0] = getLocateString(PreferencesState.getInstance().getLanguageCode(), R.string.font_extra_small);
+        newEntries[1] = getLocateString(PreferencesState.getInstance().getLanguageCode(), R.string.font_small);
+        newEntries[2] = getLocateString(PreferencesState.getInstance().getLanguageCode(), R.string.font_medium);
+        newEntries[3] = getLocateString(PreferencesState.getInstance().getLanguageCode(), R.string.font_large);
+        newEntries[4] = getLocateString(PreferencesState.getInstance().getLanguageCode(), R.string.font_extra_large);
+
+        listPreference.setEntries(newEntries);
```
xurxodev: @idelcano entries is assigned twice, one in this method and two in loadFontStyleListPreference method. loadFontStyleListPreference was created because entries and entries values for fontsizes preferences should be create from font style.
In loadFontStyleListPreference method entries translated text is not working, seems to always get the strings of English resources. I think that solution should be to do working loadFontStyleListPreference with translations and not add them twice. Can you review, please?

PR #1129, opened 2017-04-26
Not saving the warning values on the database

### LINKED ISSUE #1113: Trying to push warning and showing error

manuelplazaspalacio:
Close #1113


xurxodev:
@manuelplazaspalacio after talked with @ifoche we think that is better option save warning values and filter in push process as we are doing currently with reminder and other question types. In ConvertToSDKVisitor there are this condition for convert values value.getQuestion().hasDataElement() and hasDataElement method read from an array in constants with types that there are not upload to server.

manuelplazaspalacio:
Close #1113 @xurxodev

PR #1200, opened 2017-05-12
Bug remove user surveys logout

### LINKED ISSUE #1198: When I do logout from manu user and do login with ifoche, surveys from manu exists in the app

manuelplazaspalacio:
Close #1198


manuelplazaspalacio:
@xurxodev I think it has to be done in the login process because:
- If put the app in the background automatically the app do the logout, and If you have saved surveys yo will lose it.
- If I save a preference when the user do manual logout to delete the surveys as you are saying, If the app do the automatic logout and then other user do the login, the previous user surveys will apear in the app.

xurxodev:
@manuelplazaspalacio ok I think I have explained myself wrong:
I think that delete survey is a logout subprocess and is not login subprocess funcionally and should be executed from logout usecase. 
We have two scenarios as goal.
- automatic logout should not execute logout really because if user login again lost data
- manual logout should execute logout and remove data (included surveys)
My proposal is:
- Automatic logout not execute logout, finish all activities and save a preference AutomaticLogout = true, when app back to foreground in login activity verify this preference for navigate to dashboard or not.
- Manual logout execute logout normally 
- when user do login if AutomaticLogout = true and credentials is diferents of stored credentials than execute logout use case before login action
- Logout use case always remove use case
My intention is not mix subprocess between login and logout use case for avoid future problems

manuelplazaspalacio:
Changes request by @xurxodev done.

app/src/ereferrals/java/org/eyeseetea/malariacare/data/database/utils/PopulateDBStrategy.java
```diff
@@ -59,7 +62,12 @@ public InputStream openFile(Context context, String table) throws IOException,
 
     @Override
     public void logoutWipe() {
-
+        List<Survey> surveys = Survey.getAllSurveys();
+        for (Survey survey : surveys) {
+            survey.delete();
+        }
+        ICredentialsRepository credentialsLocalDataSource = new CredentialsLocalDataSource();
```
xurxodev: @manuelplazaspalacio this class is only for wipe db, can you move clean up ereferrals credentials to AuthenticationLocalDataSourceStrategy as normal credentials for more readable code? And instead of save null credentials better if we create a clearCredentials method as in AuthenticationLocalDataSource for has more readable code.

manuelplazaspalacio:
Changes request by @xurxodev done.

PR #1229, opened 2017-05-17
Feature survey json converter

### LINKED ISSUE #1151: Implement new push/push answer json structure

manuelplazaspalacio:
Close #1151


app/src/main/java/org/eyeseetea/malariacare/domain/usecase/push/PushUseCase.java
```diff
@@ -75,50 +78,13 @@ public void execute(final Callback callback) {
...
-                        } else {
-                            notifyReOpenOrgUnit();
-                        }
-                    }
-                });
+        mPushUseCaseStrategy.run();
     }
 
-    private boolean isOrgUnitBanned() {
+    public boolean isOrgUnitBanned() throws NetworkException, ApiCallException {
```
xurxodev: is dangerous refactor to public this methods for be invokes from strategy but can be invoked from presentation layer also, in PushUseCase only execute method should be public then I think is better refactor for move this methods to strategy base

app/src/main/java/org/eyeseetea/malariacare/domain/usecase/push/PushUseCase.java
```diff
@@ -44,7 +45,7 @@
         void onApiCallError(ApiCallException e);
     }
 
-    private IPushController mPushController;
+    public IPushController mPushController;
```
xurxodev: is dangerous refactor to public pushController for be invokes from strategy but can be invoked from presentation layer also, in PushUseCase only execute method should be public then I think is better refactor for pass pushcontroller as argument in strategy constructor

app/src/main/java/org/eyeseetea/malariacare/domain/usecase/push/PushUseCase.java
```diff
@@ -196,6 +162,14 @@ private void banOrgUnit() {
         }
     }
 
+    public IPushController getPushController() {
```
xurxodev: If we pass PushController and OrganisationUnitRepository as arguments to strategy constructor this getters is not necessary

manuelplazaspalacio:
@xurxodev Changes done.

manuelplazaspalacio:
@xurxodev  put the needed methods in the PushUseCaseStartegy and removed it from the PushUseCase

PR #1236, opened 2017-05-18
Feature refactor review fragment

### LINKED ISSUE #1163: Refactor ReviewFragment preparevalues using model view presenter

idelcano:
Closes #1163 
Myanmar reviewfragment is not working, but the bug is in development branch


app/src/main/java/org/eyeseetea/malariacare/presenter/ReviewPresenter.java
```diff
@@ -0,0 +1,122 @@
...
+                    reviewValues.add(value);
+                }
+            }
+        }
+        return reviewValues;
+    }
+
+    @Override
+    public void onClickOnValue(String UId) {
+        DashboardActivity.dashboardActivity.hideReview(UId);
```
xurxodev: Presenter should not invoke directly to presentation object. Only can invoke methods of ReviewView to communicate with presentation layer. Can you add navigateToQuestion(String uid) to ReviewView?

app/src/main/java/org/eyeseetea/malariacare/presenter/ReviewPresenter.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+        if (colorsList.size() == 1 && values.size() > 1) {
+            colorsList.add("#9c7f9b");
+        }
+        return colorsList;
+    }
+
+    private List<org.eyeseetea.malariacare.data.database.model.Value> orderValues(
+            List<org.eyeseetea.malariacare.data.database.model.Value> values) {
+        return ReviewFragmentStrategy.orderValues(values);
```
xurxodev: Presenter should not invoke directly to presentation object. Only can invoke methods of ReviewView to communicate with presentation layer. I think that order should be domain responsibility, can you refactor this code for order in GetReviewValuesBySurveyIdUseCase and create strategies by variant if its necessary?

app/src/main/java/org/eyeseetea/malariacare/presenter/ReviewPresenter.java
```diff
@@ -0,0 +1,122 @@
...
+        this.view = reviewView;
+        view.showValues(prepareValues());
+        view.initListView();
+    }
+
+    private List<org.eyeseetea.malariacare.domain.entity.Value> prepareValues() {
+        Iterator<String> colorIterator;
+        List<org.eyeseetea.malariacare.domain.entity.Value> preparedValues = new ArrayList<>();
+        List<org.eyeseetea.malariacare.data.database.model.Value> values;
+        values = getReviewValues();
```
xurxodev: Presenter only should know domain entities and not data models, can you create a GetReviewValuesBySurveyIdUseCase that returns a list of domain Value and inject this new use case in constructor?

manuelplazaspalacio:
Close #1163

app/src/main/java/org/eyeseetea/malariacare/domain/usecase/AGetReviewValuesBySurveyIdUseCaseStrategy.java
```diff
@@ -0,0 +1,81 @@
...
+    public void execute(Callback callback, long surveyId) {
+        mCallback = callback;
+        mSurveyId = surveyId;
+        mAsyncExecutor.run(this);
+    }
+
+    @Override
+    public void run() {
+        List<Value> values = mIValueRepository.getValuesFromSurvey(mSurveyId);
+        Iterator<String> colorIterator = Iterables.cycle(
```
xurxodev: The logic to create a unique color list that appears on review screen is not a domain logic or presentation logic, it´s a painting logic or view logic then we should move it to activity, fragment ...

app/src/main/java/org/eyeseetea/malariacare/layout/adapters/dashboard/ReviewScreenAdapter.java
```diff
@@ -112,7 +117,13 @@ public View getView(final int position, View convertView, ViewGroup parent) {
 
         TableRow rowView = (TableRow) this.lInflater.inflate(getRecordLayout(), parent, false);
 
-        ReviewFragmentStrategy reviewFragmentStrategy = new ReviewFragmentStrategy();
+        ReviewFragmentStrategy reviewFragmentStrategy =
```
xurxodev: Really ReviewFragmentStrategy is not a strategy for ReviewFragment is a strategy for ReviewScreenAdapter, can you rename to ReviewScreenAdapterStrategy

app/src/main/java/org/eyeseetea/malariacare/presenter/ReviewPresenter.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    ReviewView view;
+
+    public ReviewPresenter() {
+    }
+
+    public void attachView(ReviewView reviewView) {
+        this.view = reviewView;
+        IMainExecutor mainExecutor = new UIThreadExecutor();
```
xurxodev: The dependencies should not be created from the presenter, but from the composition root, as we do not have Dagger is still better to be created and assigned from the view (activity or fragment)

app/src/main/java/org/eyeseetea/malariacare/presenter/ReviewPresenter.java
```diff
@@ -0,0 +1,55 @@
...
+        void showValues(List<Value> values);
+
+        void initListView();
+
+        void navigateToQuestion(String uId);
+    }
+
+    ReviewView view;
+
+    public ReviewPresenter() {
```
xurxodev: better if we inject UseCase dependency in the constructor as parameter

app/src/cambodia/java/org/eyeseetea/malariacare/domain/usecase/strategies/GetReviewValuesBySurveyIdUseCaseStrategy.java
```diff
@@ -0,0 +1,46 @@
...
+import org.eyeseetea.malariacare.domain.boundary.repositories.IValueRepository;
+import org.eyeseetea.malariacare.domain.entity.Value;
+import org.eyeseetea.malariacare.domain.usecase.AGetReviewValuesBySurveyIdUseCaseStrategy;
+import org.eyeseetea.malariacare.layout.adapters.survey.navigation.NavigationController;
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+public class GetReviewValuesBySurveyIdUseCaseStrategy extends
```
xurxodev: I think that we have a mix of nomenclature about when we assign a name as strategy and when not. I think that root element should be named without strategy in name as LoginActivity, LoginUseCase and then if this root element has little different behaviors between variants then create strategies as ALoginActivityStrategy, IPullDhisSDKDataSourceStrategy that it's invoked from root element but if the behavior difference is big then create a base abstract class as aLoginUseCase in main if there are common code and derivate class in every variant as LoginUseCase but root element should have not strategy in name. Can you refactor this use case according to this? thanks

manuelplazaspalacio:
@xurxodev Request changes done.

PR #1273, opened 2017-05-23
Bug wevview offline beahaviour

### LINKED ISSUE #1252: Offline behavior of webviews must be revamped
When app goes offline, currently it is displaying the typical error page of a browser (with the username and pass encoded in the URL that's not good). Instead of that, when app goes offline the behavior should be:
1) If a webview have already been loaded, just keep it as-is
2) If a webview was not loaded, instead of the error page it should display a message saying "You're offline, to display this information please go online".

manuelplazaspalacio:
Close #1252


manuelplazaspalacio:
@ifoche Changes done.

PR #1305, opened 2017-05-30
Bug validation of child questions

### LINKED ISSUE #1303: Once telephone mask doesn't pass, it's impossible to send a survey without phone
If you enter a survey with an invalid phone, and then decide to change to no phone, it simply doesn't allows you to send the survey, just like it were still not validating that field (that is no longer required)

idelcano:
Closes #1303


app/src/main/java/org/eyeseetea/malariacare/views/question/IValidationQuestion.java
```diff
@@ -0,0 +1,7 @@
+package org.eyeseetea.malariacare.views.question;
+
+public interface IValidationQuestion {
```
xurxodev: this interface only is used in a base class CommonQuestionView, then I think is not necessary

app/src/main/java/org/eyeseetea/malariacare/views/question/CommonQuestionView.java
```diff
@@ -0,0 +1,34 @@
...
+        super(context, attrs);
+    }
+
+    public CommonQuestionView(Context context,
+            @Nullable AttributeSet attrs, int defStyleAttr) {
+        super(context, attrs, defStyleAttr);
+    }
+
+    @Override
+    public boolean isActive() {
```
xurxodev: default value for isActive is false, this is dangerous because only is setting when is a child question. I think default value to true is better

app/src/main/java/org/eyeseetea/malariacare/layout/adapters/survey/DynamicTabAdapter.java
```diff
@@ -820,10 +822,17 @@ private boolean toggleChild(TableRow row, Question rowQuestion, Question childQu
             Survey survey = SurveyFragmentStrategy.getSessionSurveyByQuestion(rowQuestion);
 
             if (rowQuestion.isHiddenBySurveyAndHeader(survey)) {
+                row.clearFocus();
                 row.setVisibility(View.GONE);
+                ((CommonQuestionView) row.getChildAt(0)).setActive(false);
+                Object inputView = ((AKeyboardQuestionView) row.getChildAt(0)).getInputView();
```
xurxodev: I dont´t like this code, if a question is not active then inside it should remove input error, seems this responsibility is of questionview not dynamictabadapter

app/src/main/java/org/eyeseetea/malariacare/views/question/AKeyboardQuestionView.java
```diff
@@ -43,4 +43,6 @@ protected void hideKeyboard(View view) {
     public interface onAnswerChangedListener {
         void onAnswerChanged(View view, String newValue);
     }
+
+    public abstract EditText getInputView();
```
xurxodev: I think is not a good idea expose input view to outside, what ever logic with input view child should be resolved in its question view parent

idelcano:
@xurxodev done

PR #1363, opened 2017-08-07
Feature create uid generator

### LINKED ISSUE #1339: Create UID generator
- Same approach than “the epoch” (seconds since 1-1-1970) but starting on 1-1-2017. 
- To obtain only 10 digits, instead of miliseconds we will use deciseconds starting in 1000000000 to 9999999999.
- Only activated if “no Mobile” option has been selected
- Show voucher number in the list

manuelplazaspalacio:
Close #1339


app/src/ereferrals/java/org/eyeseetea/malariacare/strategies/DashboardActivityStrategy.java
```diff
@@ -82,7 +83,12 @@ public void newSurvey(Activity activity) {
 
     @Override
     public void sendSurvey() {
-        Session.getMalariaSurveyDB().updateSurveyStatus();
+        SurveyDB malariaSurvey = Session.getMalariaSurveyDB();
+        malariaSurvey.updateSurveyStatus();
+        if (malariaSurvey.isCompleted()) {
+            malariaSurvey.setEventUid(String.valueOf(UIDGenerator.generateUID()));
```
xurxodev: UID generation should be encapsulated inside domain entity, We should not be able to assign uid from outside of domain entity

app/src/ereferrals/java/org/eyeseetea/malariacare/domain/entity/UIDGenerator.java
```diff
@@ -0,0 +1,16 @@
+package org.eyeseetea.malariacare.domain.entity;
+
+import java.util.Date;
+
+public class UIDGenerator {
```
xurxodev: we should create unit tests that verifying business rules with uid according to the client

app/src/ereferrals/java/org/eyeseetea/malariacare/strategies/DashboardActivityStrategy.java
manuelplazaspalacio: We have to change a lot of files to encapsule it into  a domain entity, because Session returns DB surveys.

manuelplazaspalacio:
Request changes by @xurxodev done.

PR #1367, opened 2017-08-09
Feature make background push active on closed app

### LINKED ISSUE #1350: Make app push work permanently
The goal is to modify the background service that implements the push so it can do it permanently even if the app is closed.

We have a few consequences in foreground depending on the push result that we probably would need to change the way they're working (to avoid crashes)

idelcano:
CLoses #1350


app/src/ereferrals/java/org/eyeseetea/malariacare/strategies/BaseActivityStrategy.java
```diff
@@ -178,7 +178,6 @@ public void onCreate() {
     }
 
     public void logout() {
-        AlarmPushReceiver.cancelPushAlarm(mBaseActivity);
```
xurxodev: Remove this code mean that when user do logout, the push will execute for ever? I think that logout should cancel the push process

app/src/main/java/org/eyeseetea/malariacare/BaseActivity.java
```diff
@@ -96,8 +95,23 @@ protected void onCreate(Bundle savedInstanceState) {
             surveyDB.setStatus(Constants.SURVEY_QUARANTINE);
             surveyDB.save();
         }
-        alarmPush = new AlarmPushReceiver();
-        alarmPush.setPushAlarm(this);
+        AlarmPushReceiver.setPushAlarm(this);
```
xurxodev: this code is very strange, can you explain to me because is necessary?

app/src/main/java/org/eyeseetea/malariacare/domain/exception/NullContextException.java
```diff
@@ -0,0 +1,17 @@
+package org.eyeseetea.malariacare.domain.exception;
+
+
+public class NullContextException extends Exception {
```
xurxodev: We should create domain exception when the problem is an expected and functional exception but if context is null is a developer error, I think that we don´t need a NullContextException.

app/src/main/java/org/eyeseetea/malariacare/domain/usecase/push/PushUseCase.java
```diff
@@ -41,6 +35,8 @@
         void onApiCallError();
 
         void onApiCallError(ApiCallException e);
+
+        void onNullContext();
```
xurxodev: UseCase should invoke callbacks with domain concepts in name, null and context are not domain concepts

app/src/main/java/org/eyeseetea/malariacare/receivers/AlarmPushReceiver.java
```diff
@@ -40,17 +40,7 @@
 
     //TODO: period has to be parameterized
     private static final long SECONDS = 1000;
-
-    public static void cancelPushAlarm(Context context) {
```
xurxodev: there are situations where we should cancel the push process as logout action, now cancel push process never is executed and I think is a problem

PR #1375, opened 2017-08-11
Feature add drive initial workflow

### LINKED ISSUE #1347: Introduce Google Drive API integration

idelcano:
Closes #1347 
(1/2)


app/src/main/java/org/eyeseetea/malariacare/data/remote/drive/DriveRestController.java
```diff
@@ -0,0 +1,192 @@
...
+import org.eyeseetea.malariacare.R;
+import org.eyeseetea.malariacare.data.database.model.MediaDB;
+import org.eyeseetea.malariacare.data.database.utils.PreferencesState;
+import org.eyeseetea.malariacare.data.repositories.MediaRepository;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Arrays;
+
+public class DriveRestController {
```
xurxodev: this class seems a DownloadMediaUseCase, can you refactor this class to use case respecting clean architecture rules abstracting external dependencies?

app/src/main/java/org/eyeseetea/malariacare/domain/usecase/pull/DownloadMediaUseCase.java
```diff
@@ -0,0 +1,259 @@
...
+                    R.raw.driveserviceprivatekey);
+            serviceCredential = GoogleCredential.fromStream(privateJsonStream).createScoped(
+                    Arrays.asList(SCOPES));
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
+    }
+
+    private boolean isGooglePlayAppAvailable() {
```
xurxodev: this is a Google Drive service requirement and not a domain requirement then this validation should be realized in new FileDownloader adapter to create.

app/src/main/java/org/eyeseetea/malariacare/domain/usecase/pull/DownloadMediaUseCase.java
```diff
@@ -0,0 +1,259 @@
...
+        Dialog dialog = apiAvailability.getErrorDialog(
+                DashboardActivity.dashboardActivity,
+                connectionStatusCode,
+                REQUEST_GOOGLE_PLAY_SERVICES);
+        dialog.show();
+    }
+
+
+
+    private void showMessage(String message) {
```
xurxodev: use case can not show messages directly, this should be realized with callbacks

app/src/main/java/org/eyeseetea/malariacare/domain/usecase/pull/DownloadMediaUseCase.java
```diff
@@ -0,0 +1,259 @@
...
+            return false;
+        }
+    }
+
+    /**
+     * Checks whether the device currently has a network connection.
+     *
+     * @return true if the device has a network connection, false otherwise.
+     */
+    private boolean isDeviceOnline() {
```
xurxodev: Download files only when wifi is enabled is a domain requirement but UseCase can not invoke external layer directly, we need to create a port interface inside domain layer as with the interfaces of repositories and then we need to create the adapter that communicates with the external layer as repositories. In this case for example:
Interface: IConnectivityManager inside domain layer and ConnectivityManager inside network layer and this dependency are injected it through the constructor as repositories.
UseCase should not know the Android framework component used for verifying connectivity then if in the future the framework change the use case is not affected

app/src/main/java/org/eyeseetea/malariacare/fragments/AVFragment.java
```diff
@@ -96,21 +95,7 @@ private void loadMediaListAndAdapter() {
         //MediaRepository mediaRepository = new MediaRepository();
 
         //TODO: remove fake media repository
```
xurxodev: please merge #1366 PR with this PR because has changes in this part

app/src/main/java/org/eyeseetea/malariacare/DashboardActivity.java
```diff
@@ -763,6 +764,19 @@ public void onTabChanged(String tabId) {
         }
     }
 
+    /**
+     * Handles resolution callbacks.
+     */
+    @Override
+    protected void onActivityResult(int requestCode, int resultCode,
```
xurxodev: Why onActivityResult is necessary if we are not using startActivityForResult() method?

app/src/main/java/org/eyeseetea/malariacare/domain/entity/Media.java
```diff
@@ -58,11 +69,27 @@ public static String getFileSize(){
         return "0MB";
```
xurxodev: Why returns this hard coded result?

app/src/main/java/org/eyeseetea/malariacare/DashboardActivity.java
idelcano: @xurxodev  It is called here: https://github.com/EyeSeeTea/pictureapp/pull/1375/files#diff-93401c54e92d24bdf56298493f2c7799R93

app/src/ereferrals/java/org/eyeseetea/malariacare/strategies/DashboardActivityStrategy.java
```diff
@@ -15,20 +23,45 @@
...
+import org.eyeseetea.malariacare.presentation.executors.AsyncExecutor;
 
+import java.util.HashMap;
 
 public class DashboardActivityStrategy extends ADashboardActivityStrategy {
+    public static final int REQUEST_GOOGLE_PLAY_SERVICES = 102;
+
+    static final int REQUEST_AUTHORIZATION = 101;
+
+    public interface Callback {
```
xurxodev: DashboardActivityStrategy should not define the callback, this should be define inside of use case and has no external dependencies for example:

public interface Callback{
  void onError(FileDownloadException ex);
  void onSuccess(HashMap<String, String> syncedFiles);
}

we can create FileDownloadException for group problems and use cause property to send encapsulated exceptions as UserRecoverableAuthIOException, GooglePlayServicesAvailabilityIOException ...

public class FileDownloadException extends Exception{
     public FileDownloadException(String message, Throwable cause) {
        super(message, cause);
    }
}

app/src/ereferrals/java/org/eyeseetea/malariacare/strategies/DashboardActivityStrategy.java
```diff
@@ -39,6 +72,83 @@ public void onCreate() {
...
+                        if (correctSyncedFiles > 0) {
+                            showToast(String.format("%d files synced", correctSyncedFiles));
+                        }
+                    }
+
+                    @Override
+                    public void onRemove(String uid) {
+                        System.out.println(
+                                "downloadFile error (file extension not supported)" + uid);
+                        mediaRepository.deleteModel(mediaRepository.findByUid(uid));
```
xurxodev: a class of presentation layer should not invoke mediaRepository, this is a use case responsibility

PR #1379, opened 2017-08-14
refactor instrumentation test with MockWebServer to unit test against…

### LINKED ISSUE #1373: Review ForgotPassword test with MockWebServer

xurxodev:
… eReferralsAPIClient
Closes #1373


xurxodev:
@ifoche and @manuelplazaspalacio 
@ifoche @manuelplazaspalacio  After reviewing the tests with MockWebServer I found some problems:
- The preference for the URL of the web service was being changed in the tests and this caused it to be changed in the application causing errors on runtime
- The use case was tested using instrumentation tests but I think it is better if we test the api client directly using unit tests and mock web server, so we have a smaller test scope and in this current architecture the tests are less brittle with app code refactors
For the future:
- An espresso tests mocking dependencies are not possible at this moment because there are many dependencies are not inject it in all code and we are not using dagger for injecting it from outside activities and fragments

ifoche (approved):
thanks @xurxodev

PR #1392, opened 2017-08-18
Feature introduce language api calls

### LINKED ISSUE #1374: Introduce language in API calls
Language must be included in WS API calls payload with the following format:

	"language": "en",

at the top level of the json. 

The trick with the language would be that:
1. If the settings say "system default" for the language configuration, we need to get the language from the system
2. Otherwise, the code must be the one selected in settings

manuelplazaspalacio:
Close #1374


app/src/ereferrals/java/org/eyeseetea/malariacare/domain/boundary/repositories/ICurrentLanguageRepository.java
```diff
@@ -0,0 +1,7 @@
+package org.eyeseetea.malariacare.domain.boundary.repositories;
+
+import org.eyeseetea.malariacare.data.IDataSourceCallback;
+
+public interface ICurrentLanguageRepository {
```
xurxodev: Repositories should return domain entities, this Repository returns a primitive. Here it may make sense a Settings domain entity with CurrenLanguage getter in this moment and refactor this repository to ISettingsRepository

app/src/ereferrals/java/org/eyeseetea/malariacare/domain/boundary/repositories/ICurrentLanguageRepository.java
```diff
@@ -0,0 +1,7 @@
...
+    void getCurrentLanguage(IDataSourceCallback<String> callback);
```
xurxodev: This method retrieves data from preferences and is not necessary return data invoking callback, can you refactor for return domain entity directly?

manuelplazaspalacio:
Changes requested by @xurxodev done.

PR #1396, opened 2017-08-22
Cosmetics move start button as floating button

### LINKED ISSUE #1384: Make + button a floating button
+ button must follow the scroll and be always present in the form of a floating button, to avoid forcing the users to scroll down to the end of the list to create a voucher

idelcano:
Closes #1384 
This feature is not tested in lao and cambodia, but should works


idelcano:
@xurxodev  I updated the ereferrals variant with floatingbutton and the new style.
In all the variants the button is always visible but i couldn't test myanmar and laos.
I couldn't test the initfooter feature but it should work now.
The other variants still having a imagebutton, because the floatingbutton style is different and i think we need decide about the button style in the next releases, the button style is not different than the last release and the functionality is added using a imagebutton.

PR #1408, opened 2017-08-24
set quarantine surveys as completed for ereferrals

### LINKED ISSUE #1388: Quarantine surveys not pushing
If there are quarantine surveys they will never push.
I was putting app offline and at any point the app put the surveys in quarantine, I think it was in the on create method of the BaseActivity.
I think to solve this bug you only need to try to push the surveys in quarantine with the completed surveys in the push method of the WSPushController.

idelcano:
Closes #1388


app/src/main/java/org/eyeseetea/malariacare/data/database/datasources/SurveyLocalDataSource.java
```diff
@@ -44,4 +45,13 @@ public void getUnsentSurveys(IDataSourceCallback<List<Survey>> callback) {
         }
         callback.onSuccess(unsentSurveys);
     }
+
+    @Override
+    public void updateAllQuarantineSurveysAsComplete() {
```
xurxodev: the responsibilities of a repository is getting and save in database and map from db model to domain entity but have no responsibility for change survey status. I think that this responsibility is a domain responsibility then the repository should has a GetAllQuantineSurveys method and return domain survey entities then the PushUseCase iterate surveys and change status and invoke save method in the repository.

app/src/ereferrals/java/org/eyeseetea/malariacare/services/strategies/PushServiceStrategy.java
```diff
@@ -176,7 +175,7 @@ protected void executePush() {
                 new PushUseCase(pushController, asyncExecutor, mainExecutor,
                         surveysThresholds, surveyRepository, orgUnitRepository);
 
-        SurveyChecker.launchQuarantineChecker();
+        surveyRepository.updateAllQuarantineSurveysAsComplete();
```
xurxodev: the decision of update quarantine surveys to complete is a domain decision, not a presentation decision then the Push use case should realize this decision.

PR #1409, opened 2017-08-24
Feature add progress bar on av fragment

### LINKED ISSUE #1405: Show a message on media tab when media is syncing

idelcano:
Closes #1405


app/src/main/java/org/eyeseetea/malariacare/fragments/AVFragment.java
```diff
@@ -78,15 +82,15 @@ public void onDestroy() {
     }
 
     private void initializePresenter() {
-        IMainExecutor mainExecutor = new UIThreadExecutor();
```
xurxodev: MainExecutor should be executed only inside use cases for notify callbacks

app/src/main/java/org/eyeseetea/malariacare/fragments/AVFragment.java
```diff
@@ -152,4 +156,17 @@ public void onClick(Media media) {
     public void reloadData() {
         initializePresenter();
     }
+
+    public void showProgress() {
+        mainExecutor.run(new Runnable() {
```
xurxodev: MainExecutor should be executed only inside use cases for notify callbacks

app/src/main/java/org/eyeseetea/malariacare/fragments/AVFragment.java
```diff
@@ -152,4 +156,17 @@ public void onClick(Media media) {
...
+            @Override
+            public void run() {
+                if(PreferencesState.getInstance().isFileDownloaderInProgress()) {
```
xurxodev: AvFragment should not invoke PreferenceState because is in data layer, better if according to current code we create two methods showDownloadProgress and hideDownloadProgress to invoke from DashboardActivityStrategy

app/src/main/java/org/eyeseetea/malariacare/domain/usecase/pull/DownloadMediaUseCase.java
```diff
@@ -68,6 +70,11 @@ public void onError(FileDownloadException ex) {
                         }
 
                         @Override
+                        public void onProgress() {
+                            mCallback.onProgress();
```
xurxodev: callback invocations should be executed in main thread using MainExecutor as other use cases

app/src/ereferrals/java/org/eyeseetea/malariacare/strategies/DashboardActivityStrategy.java
```diff
@@ -352,6 +352,11 @@ public void onSuccess(int syncedFiles) {
                     showToast(String.format("%d files synced", syncedFiles));
                 }
             }
+
+            @Override
+            public void onProgress() {
```
xurxodev: this callback method should receive a parameter for indicate ProgressStatus as begin or end

PR #1420, opened 2017-09-11
Added WSPush timeout

### LINKED ISSUE #1413: Introduce a variable timeout for WS push
WS can spend some variable time to answer a push call. The suggested function that would describe the timeout to be configured in the call is: number_vouchers x 2000ms

idelcano:
Closes #1413


xurxodev (approved):
thanks @idelcano

PR #1423, opened 2017-09-11
Feature reassing uid to survey

### LINKED ISSUE #1401: Reassign UID depending on server answer to survey push
If UID exists in the server, the reply to user push will contain a new UID for that survey, that will be changed in the DB and logged in logcat

manuelplazaspalacio:
Close #1401


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1424, opened 2017-09-11
Feature manage duplicate surveys

### LINKED ISSUE #1402: Error treatment for duplicated survey
In case a pushing survey is duplicated server will answer with an error and we need to parse it and show it on logcat. Else, that survey will be moved to SENT state

manuelplazaspalacio:
Close #1402


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1425, opened 2017-09-11
Feature prevent logout on server ou pin change

### LINKED ISSUE #1418: Check login logic
- [x] On OU password change. Does that produce a logout?

idelcano:
Closes #1418 
I modify a push exception to prevent the logout, 
and prevent a logout if the ou pin is changed on sever.

But i had some problems with clone server and this branch not related with this issue( i can't create surveys, and on login a exception is shown "Error while trying to pull metadata from dhis server")


app/src/ereferrals/java/org/eyeseetea/malariacare/data/sync/exporter/WSClient.java
```diff
@@ -78,8 +79,8 @@ public void pushSurveys(SurveyContainerWSObject surveyContainerWSObject,
             ConversionException conversionException = new ConversionException(null);
             wsClientCallBack.onError(conversionException);
         } else if (response != null && response.code() == 402) {
-            InvalidCredentialsException invalidCredentialsException =
-                    new InvalidCredentialsException();
+            InvalidCreedentialsOnPushException invalidCredentialsException =
```
xurxodev: It is not necessary to create the new InvalidCreedentialsOnPushException because this error is with HardCoded credentials, InvalidCredentialsException here does not provoke Logout

app/src/main/java/org/eyeseetea/malariacare/domain/exception/InvalidCreedentialsOnPushException.java
```diff
@@ -0,0 +1,10 @@
+package org.eyeseetea.malariacare.domain.exception;
+
+public class InvalidCreedentialsOnPushException extends Exception {
```
xurxodev: It is not necessary to create the new InvalidCreedentialsOnPushException

app/src/ereferrals/java/org/eyeseetea/malariacare/domain/usecase/LoginUseCase.java
```diff
@@ -166,7 +165,15 @@ private void checkUserCredentialsWithOrgUnit(Credentials credentials,
             if (fromNetWorkError) {
                 notifyNetworkError();
             } else {
-                notifyInvalidCredentials();
+                if (insertedCredentials.getUsername().equals(credentials.getUsername())
```
xurxodev, line 168: if on the remote server pin for a user is changed then in a push process when validate username and pin previously to the push process,  the app should navigate to login without to do logout but with this code push is blocking but does not navigate to login screen.

xurxodev (changes requested):
@idelcano @manuelplazaspalacio I have requested you some changes, can you review it please?

xurxodev (approved):
thanks @idelcano

PR #1463, opened 2017-09-14
loseSolve bug with login enable when no enter any text

### LINKED ISSUE #1462: Disable login button when any of username or pass is empty
It's causing a crash

manuelplazaspalacio:
Close #1462


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1476, opened 2017-09-15
add onBackPressed on webviews

### LINKED ISSUE #1455: Back button is captured by the app but not by the webview
If webview can navigate back, it should do it, and when there's no more pages to go back, do the android back

idelcano:
Closes #1455


xurxodev (approved):
thanks @idelcano I think that this implementation is correct but ereferrals web has a bug for back to the previous page from new Voucher page

PR #1478, opened 2017-09-15
avoid to do changes in uid and eventDate when completed surveys are o…

### LINKED ISSUE #1477: When open and send a completed and non sent survey should not change uid and event date

xurxodev:
…pened and sent from review screen
closes #1477


ifoche (approved):
thanks @xurxodev

PR #1502, opened 2017-09-18
Show drive folder in settings only if developer option active

### LINKED ISSUE #1494: Hide drive folder option under "developers options"
Just like with the export DB, don't show the "drive folder" setting unless developers options is activated

manuelplazaspalacio:
Close #1494


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1503, opened 2017-09-18
Solved bug showing error when pushing surveys

### LINKED ISSUE #1496: MZ issuing error
This is the alpha-tester report:

I receive the following error:
"An error has occurred to the conversion in push process"

So the voucher stays on the queue.
But I actually received a voucher number via SMS.

The voucher gets sent again and I receive a SMS again, and again, and...

@manuelplazaspalacio can you please take a look to what's returning the WS and guess why is this happening?

manuelplazaspalacio:
Close #1496


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1505, opened 2017-09-18
Solved bug not hiding childs of childs when changing parebt option in…

### LINKED ISSUE #1501: Address not deleted when current method's changed
Alpha tester report:

Example from Mozambique
1. I enter "Current Method" = "Pills"
2. "HH delivery" shows and I enter "yes"
3. "Address" shows and I enter "anything"
4. I switch "Current Method" to "Condom"
5. "HH delivery" hides but not "Address"

I actually want "Address" to be set empty and hidden again

manuelplazaspalacio:
Close #1501


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1510, opened 2017-09-19
Bug review csv hidden

### LINKED ISSUE #1495: Review CSVs taking new visibility tag into account for the config file
We need to review and update the CSV files. The new config files will be from now on tracked in [this git repo](https://github.com/BRaimbault/Connect_config). When reviewing the CSV files, take into account that there is a new tag called visibility="hidden". If something is tagged with that, we must ignore it for our CSVs

manuelplazaspalacio:
Close #1495


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1513, opened 2017-09-20
Bug fix cnm variant compilation

### LINKED ISSUE #1506: Fix cnm variant compilation

manuelplazaspalacio:
Close #1506


app/src/cnm/java/org/eyeseetea/malariacare/domain/usecase/LoginUseCase.java
```diff
@@ -24,7 +24,7 @@ public void execute(final Credentials credentials, final Callback loginCallback)
                 new IAuthenticationManager.Callback<UserAccount>() {
                     @Override
                     public void onSuccess(UserAccount userAccount) {
-                        if (!credentials.isDemoCredentials()) {
+                        if (credentials == null && !credentials.isDemoCredentials()) {
```
xurxodev: if credentials is null on execute we should throw IllegalArgumentException, so validate null onSuccess is not necesary

app/src/cnm/java/org/eyeseetea/malariacare/services/strategies/PushServiceStrategy.java
```diff
@@ -17,7 +17,7 @@ public PushServiceStrategy(PushService pushService) {
 
     @Override
     public void push() {
-        if (Session.getCredentials().isDemoCredentials()) {
+        if (Session.getCredentials() == null || Session.getCredentials().isDemoCredentials()) {
```
xurxodev: if credentials is null should not execute mocked push

app/src/main/java/org/eyeseetea/malariacare/ProgressActivity.java
```diff
@@ -125,7 +125,8 @@ public void onLogoutError(String message) {
     @Override
     public void onResume() {
         super.onResume();
-        launchPull(Session.getCredentials().isDemoCredentials());
+        launchPull(
```
xurxodev: null credentials and demo credentials is no the same case

app/src/main/java/org/eyeseetea/malariacare/data/authentication/AuthenticationManager.java
```diff
@@ -34,7 +34,7 @@ public AuthenticationManager(Context context) {
     @Override
     public void login(final Credentials credentials,
             final IAuthenticationManager.Callback<UserAccount> callback) {
-        if (credentials.isDemoCredentials()) {
+        if (credentials == null || credentials.isDemoCredentials()) {
```
xurxodev: if credentials is null on login we should throw IllegalArgumentException because null credentials and demo credentials is not the same case

app/src/main/java/org/eyeseetea/malariacare/network/ServerAPIController.java
```diff
@@ -482,7 +482,7 @@ public static UserDB pullUserAttributes(UserDB loggedUserDB) throws ApiCallExcep
     }
 
     public static boolean isUserClosed(String userUid) throws ApiCallException {
-        if (Session.getCredentials().isDemoCredentials()) {
+        if (Session.getCredentials() == null || Session.getCredentials().isDemoCredentials()) {
```
xurxodev: null credentials and demo credentials is not the same case

xurxodev (changes requested):
@manuelplazaspalacio the behavior has changed with current credentials null management. For example When user is not in demo mode and is not logged (null credentials), in these case surveys should not appears never in send list after push process, 
Can you review whole null credentials management?

manuelplazaspalacio:
Changes done @xurxodev

xurxodev (approved):
thanks @manuelplazaspalacio

PR #1524, opened 2017-09-26
Allow app move to production:

### LINKED ISSUE #1519: Allow the app to move to production
In order to move to production the app would need to:
- [x] Move "eRefWSTrain" part of the string in SurveyApiClientRetrofit to the string in settings (the Web Service URL), so the user can change it
- [x] Introduce a new setting "WebView URL" extracting from donottranslate.xml the https://apps.psi-mis.org/connect-train so the user can change it

manuelplazaspalacio:
Close #1519


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1529, opened 2017-09-29
Feature auto configuration

### LINKED ISSUE #1504: Add OU auto-configuration feature
- The new auto-configuration issue must be launched on app start but will also have a button in preferences saying "Auto-configure" under the server and OU settings, to launch it again.

- Once launched, a spinning progress bar blocks the screen with the message "Auto-configuring OU, please wait..."

- The way it works is as follows:

1) Each OU will contain in its "phone" parameter, the phone numbers of the users that can auto-configure that OU. The phones will be separated by ";" symbol in case more than one phone is configured in the OU

2) When auto-configuration is launched, the user phone number will be retrieved from the device

3) An API call will be launched to the server asking for the OU whose phone number match the user one

4) In case that OU exists, it will be configured automatically in the OU setting, and the app will be ready to send data. Otherwise, a pop-up saying "No OU found for this phone number" will appear

manuelplazaspalacio:
Close #1504


app/src/cnm/java/org/eyeseetea/malariacare/strategies/SplashActivityStrategy.java
```diff
@@ -1,16 +1,209 @@
...
+    }
+
+
+    private void showErrorAutoConfiguration() {
+        new AlertDialog.Builder(activity)
+                .setTitle(R.string.error_message)
+                .setMessage(R.string.error_auto_configuration)
+                .setPositiveButton(android.R.string.ok, new DialogInterface.OnClickListener() {
+                    public void onClick(DialogInterface dialog, int arg1) {
+                        continueWithDemoUser();
```
xurxodev: when autoconfiguration fail, the app should not continue with demo user, you must do nothing, you will be in no-login mode

xurxodev (changes requested):
@manuelplazaspalacio I have requested you some changes. please, can you review it?

manuelplazaspalacio:
Changes done @xurxodev

xurxodev (changes requested):
- when on pull process occurred either error the autoconfiguration message is shown but this message should be shown only for autoconfiguration error. I think we should create an AutoconfigurationException in domain for CNM and for avoid realize many changes we can change message parameter in onError usecase callback to Throwable then in other variants get the message and execute the same code and on CNM if Throwable is AutoconfigurationException show the correct dialog.
- if autoconfiguration is completed in push process when ban orgunit is realized an exception occurred and push fail
- When autoconfiguration fail, this is the final step the pull process, then has no sense execute pull again. 
- When autoconfiguration is complete, if I click on autoconfiguration in settings we should call to server again for get orgunit

manuelplazaspalacio:
Solved bugs @xurxodev

xurxodev (approved):
thanks @manuelplazaspalacio

PR #1544, opened 2017-10-16
Feature create uo tree question

### LINKED ISSUE #1538: Create new OU_TREE question
A new question type must be included. The way this question will work will be as follows:
- The question downloads a json from an endpoint (it will be [this type of endpoint](https://docs.dhis2.org/2.22/en/developer/html/ch01s62.html) but still creating it for us)
- After downloading it, it parse the [json](https://drive.google.com/open?id=0B7TXlu17DcAbSVJLQ25JVTh5UWM), obtaining an OU tree
- The question will be a dropdown where the user can select the OU
- When sending the value of the question, the OU "Name_Vill_E" property will be used as value, and OU coordinates will be also sent

manuelplazaspalacio:
Close #1538


app/src/cnm/java/org/eyeseetea/malariacare/data/sync/importer/strategies/PullControllerStrategy.java
```diff
@@ -49,7 +64,7 @@ public void pull(PullFilters pullFilters, IPullController.Callback callback, Con
                     throw new AutoconfigureException();
                 }
             }
-            if (organisationUnit != null || pullFilters.isDemo()) {
+            if (pullFilters.isDemo()) {
                 callback.onComplete();
             } else {
                 mPullController.pullMetada(pullFilters, callback);
```
xurxodev: when the pull is executed in autoconfigure mode, this method is executed but without parameters in pullFilters to download metadata

app/src/cnm/java/org/eyeseetea/malariacare/data/sync/importer/ConvertFromApiVisitor.java
```diff
@@ -0,0 +1,156 @@
...
+                            optionVill.setOptionAttributeDB(optionAttributeDB);
+                            villageOptions.add(optionVill);
+                        }
+                    }
+                }
+            }
+        }
+        Log.d("SAVING LIST", "saving optrionstree list");
+        saveBatch(villageOptionAttributes);
+        for (Model villageOption : villageOptions) {
```
xurxodev: Please review this loops because has no sense, for every option is assign every attribute. It is 100 million loops and performance is affected, I think that creates a map and search by keymap can to improve performance

xurxodev (changes requested):
 can you review it please?

manuelplazaspalacio:
Changes done @xurxodev

xurxodev (approved):
thanks @manuelplazaspalacio. I have fixed several problems, please review my commits and talk me if you have doubts

PR #1546, opened 2017-10-18
Feature cnm csv

### LINKED ISSUE #1540: New CSVs according to CNM questions

manuelplazaspalacio:
Close #1540


xurxodev (changes requested):
- The other question should be a multiline long text but is a simple text 
- the push fail with conflicts
- In push, the property store by of data values is always null
- the dataelement k3gpW1Vm3Ov for village is not sending in push

manuelplazaspalacio:
Solved bugs @xurxodev. Use OrgUnit Bateay Neang to test, there is a bug not solved related with server configuration when pushing; {"object":"dataElement","value":"o7mG7zUBGsM  is not a valid data element"}. This is not a bug from the app.

xurxodev (changes requested):
@manuelplazaspalacio I have found the same problems again:

- The other question should be a multiline long text but is a simple text
- the push fail with conflicts (pending of client solve dataelement conclict)
- In push, the property store by of data values is always null
- the dataelement k3gpW1Vm3Ov for village is not sending in push

manuelplazaspalacio:
Solved bugs @xurxodev

ifoche:
Push is no longer failing. There was a space in DE UID in de CSV.

ifoche:
And please execute a git pull on the sdk. There was also a problem with WARNING state not being recognized in the ImportSummary, which caused a crash during the ImportSummary parsing

xurxodev (approved):
thanks @manuelplazaspalacio

PR #1557, opened 2017-10-24
show a warning if exists more than one org unit by phone

### LINKED ISSUE #1552: When there are more one org unit with my phone number, the app should show a warning and select the first org unit

xurxodev:
close #1552


app/src/main/java/org/eyeseetea/malariacare/network/ServerAPIController.java
```diff
@@ -719,15 +722,29 @@ public static OrganisationUnit getOrgUnitByPhone(String imei) throws ApiCallExce
...
+            boolean moreThanOneOrgUnit = false;
+
             //0| >1 matches -> Error
-            if (orgUnitsArray.length() == 0 || orgUnitsArray.length() > 1) {
+            if (orgUnitsArray.length() == 0) {
                 Log.e(TAG, String.format("getOrgUnitData(%s) -> Found %d matches", imei,
                         orgUnitsArray.length()));
                 return null;
+            } else if (orgUnitsArray.length() > 1) {
+                moreThanOneOrgUnit = true;
```
ifoche: @xurxodev do we need this flag? couldn't we just ask for orgUnitsArray.length()>1 down in the condition to throw the exception?

ifoche (changes requested):
@xurxodev Just a cosmetic thing, but I prefer you to review it and tell me if I'm missing something rather than changing it myself

xurxodev:
@ifoche I have removed unnecessary flag

ifoche (approved):
thanks @xurxodev

PR #1560, opened 2017-10-24
Downloading orgUniTree metadata version and if it's a higher value de…

### LINKED ISSUE #1547: Implements metadata version logic
Verify metadata version stored in namespace of data storage in server

manuelplazaspalacio:
Close #1547


xurxodev (changes requested):
@manuelplazaspalacio Can you refactor calls in retrofit client to execute it in sync way, so we avoid execute callbacks in main thread and progress bar is not freeze

manuelplazaspalacio:
Doing api calls in background. @xurxodev

xurxodev (approved):
thanks @manuelplazaspalacio I have fixed some possible bug, if we have to download only data (for example to select org unit after login in settings) does not remove obsolete data although remote metadata version has changed on the server

PR #1562, opened 2017-10-24
Feature pull ou children

### LINKED ISSUE #1551: On OU pull, convert to OU question options only those that are children of the assigned OU
- [ ] Check that works for both auto-configured and manually configured user
- [ ] Eventually this would imply modifications in OrganisationUnitRepository and ServerAPIController classes

idelcano:
Closes #1551


xurxodev (approved):
thanks @idelcano I have created issues to refactor the code in version 2

PR #1588, opened 2017-10-26
fix pull of events

### LINKED ISSUE #1585: Events are not being downloaded when you select other than "No data"

idelcano:
Closes #1585


xurxodev (approved):
thanks @idelcano

PR #1589, opened 2017-10-26
Solved bug not allowing to reach review screen

### LINKED ISSUE #1581: scrolling forward doesn't allow you to reach the review screen (blocks in last question)

manuelplazaspalacio:
Close #1581

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/layout/adapters/survey/navigation/NavigationController.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/layout/adapters/survey/navigation/NavigationController.java
@@ -111,6 +111,9 @@ public class NavigationController {
         OptionDB currentOptionDB = currentValueDB == null ? null : currentValueDB.getOptionDB();
         //Find next node with current option
         boolean isAllowed = findNext(currentOptionDB) != null;
+        if (!isAllowed && !hasNext(null) && !hasNext(currentValueDB.getOptionDB())) {
+            return true;
+        }
         Log.d(TAG, String.format("isNextAllowed()->%b", isAllowed));
         return isAllowed;
     }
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/eyeseetea/malariacare/layout/adapters/survey/navigation/NavigationController.java
```diff
@@ -111,6 +111,9 @@ public boolean isNextAllowed() {
         OptionDB currentOptionDB = currentValueDB == null ? null : currentValueDB.getOptionDB();
         //Find next node with current option
         boolean isAllowed = findNext(currentOptionDB) != null;
+        if (!isAllowed && !hasNext(null) && !hasNext(currentValueDB.getOptionDB())) {
```
xurxodev: @manuelplazaspalacio I think that we should not change this method of NavigationController for this because we are modifying a existed method that using several apps and several scenarios and is possible that we are changing a behavior. I think we should change caller (onSwipeLeft) for this method and create new condition to know if is the last question

xurxodev (changes requested):
@manuelplazaspalacio  I have requested you some changes, can you review it?

app/src/main/java/org/eyeseetea/malariacare/layout/adapters/survey/navigation/NavigationController.java
manuelplazaspalacio: This method is only called from onSwipeLeft method of dyanmicTabAdapter. And I think if its the las question that it's the correct behaviour for all apps.

manuelplazaspalacio:
Changes done @xurxodev

xurxodev (approved):
thanks @manuelplazaspalacio

PR #1590, opened 2017-10-26
Bug imei settings

### LINKED ISSUE #1579: On settings, show IMEI
Hidden under show developers options

manuelplazaspalacio:
Close #1579

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/domain/usecase/GetDeviceUseCase.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/domain/usecase/GetDeviceUseCase.java
@@ -1,5 +1,6 @@
 package org.eyeseetea.malariacare.domain.usecase;
 
+import org.eyeseetea.malariacare.domain.boundary.executors.IAsyncExecutor;
 import org.eyeseetea.malariacare.domain.boundary.executors.IMainExecutor;
 import org.eyeseetea.malariacare.domain.boundary.repositories.IDeviceRepository;
 import org.eyeseetea.malariacare.domain.entity.Device;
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/eyeseetea/malariacare/domain/usecase/GetDeviceUseCase.java
```diff
@@ -0,0 +1,39 @@
+package org.eyeseetea.malariacare.domain.usecase;
+
+import org.eyeseetea.malariacare.domain.boundary.executors.IMainExecutor;
+import org.eyeseetea.malariacare.domain.boundary.repositories.IDeviceRepository;
+import org.eyeseetea.malariacare.domain.entity.Device;
+
+public class GetDeviceUseCase implements UseCase {
```
xurxodev, line 8: use cases should use AsyncExecutor to retrieve data from data layer and MainExecutor to notify presentation layer

xurxodev (changes requested):
@manuelplazaspalacio I have requested some changes, can you review it please?

app/src/main/java/org/eyeseetea/malariacare/domain/usecase/GetDeviceUseCase.java
manuelplazaspalacio: But the device info is not in data layer, it's an Android service.

xurxodev: well, we can do two things:
- use AsyncExecutor and MainExecutor as in other use cases for we have a standard way, this is better option for me.
- Does not use AsyncExecutor and MainExecutor because really here does not communicate with a database or remote service. 

But you are using only main executor in your solution and this option has no sense

manuelplazaspalacio:
Changes done @xurxodev

xurxodev (approved):
thanks @manuelplazaspalacio

PR #1592, opened 2017-10-26
assign correct uids to treatment and other treatment

### LINKED ISSUE #1578: On other question selected, treatment DE receives the value for other question and viceversa

xurxodev:
close #1578


ifoche (approved):
thanks @xurxodev

PR #1595, opened 2017-10-26
Solved bug with languages

### LINKED ISSUE #1573: Bring language bug fix from eReferrals
Only system defined appears, but not English

manuelplazaspalacio:
Close #1573


xurxodev (changes requested):
@manuelplazaspalacio Now with this changes when I changed to jemer language, the settings activity is refreshed but the texts are in english until back to dashboard. I think that when settings is refreshed after language change the texts should be in selected language

manuelplazaspalacio:
Solved bugs @xurxodev

xurxodev (approved):
thanks @manuelplazaspalacio

PR #1596, opened 2017-10-26
show error on ApiCallException during the autoconfig pull

### LINKED ISSUE #1582: On metadata download if there's a network problem only a toast is shown and nothing happens

idelcano:
Closes #1582


xurxodev (changes requested):
@idelcano we should show util messages to user:
With your change now with either error as a timeout, server problem, network problem the app showing the message "No OU found for this phone" that should be visible only when there is not an org unit with your phone EMEI. I think is better when other errors occurred show a error like "An error has occured autoconfiguring, please retry later again"

xurxodev (approved):
thanks @idelcano

PR #1597, opened 2017-10-26
Make clickable the question container instead of the image

### LINKED ISSUE #1580: IMAGES_X type questions listener is attached to the image and not to the container

idelcano:
Closes #1580


xurxodev (approved):
thanks @idelcano

PR #1598, opened 2017-10-26
Remove order and remove review values

### LINKED ISSUE #1577: On microscope selection only 2 results appear in review screen

idelcano:
Closes #1577


xurxodev (changes requested):
@idelcano with this change camboya variant will change its behavior. We should modify only a CNM strategy like lao and ereferrals

xurxodev (approved):
thanks @idelcano

PR #1599, opened 2017-10-26
Feature crash on imagen not loaded

### LINKED ISSUE #1574: Don't try to load other question image if no image is present

idelcano:
Closes #1574


xurxodev (approved):
thanks @idelcano

PR #1601, opened 2017-10-26
Hide other question keyboard

### LINKED ISSUE #1575: keyboard is not hidden after ok in "other" question

idelcano:
Closes #1575


xurxodev (approved):
thanks @idelcano

PR #1608, opened 2017-10-27
Feature restore control dataelements

### LINKED ISSUE #1604: Send control dataElements
We had added a new gradle variable to avoid them...now we need to restore the previous behavior back and check with the server that's working

idelcano:
Closes #1604


ifoche (approved):
thanks @idelcano

PR #1612, opened 2017-10-27
Feature bug test result not works on pulled surveys

### LINKED ISSUE #1609: Bug showing microscope result question when comes from pull

idelcano:
Closes #1609


ifoche (approved):
thanks @idelcano

PR #1631, opened 2017-11-08
Hiding options by question threshold option match.

### LINKED ISSUE #1624: Introduce show/hide for female/pregnant option
If the age is lower than 13, pregnant option should not appear

manuelplazaspalacio:
Close #1624


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1634, opened 2017-11-08
Remember the username the first time the user inserts a correct one.

### LINKED ISSUE #1620: Remember username after first login
Once user login the first time, the username should be remembered (in preferences) and for next logins, it should be pre-filled in the login screen

manuelplazaspalacio:
Close  #1620


app/src/ereferrals/java/org/eyeseetea/malariacare/strategies/LoginActivityStrategy.java
```diff
@@ -127,6 +128,19 @@ public void onClick(View view) {
                 onForgotPassword();
             }
         });
+        IMainExecutor mainExecutor = new UIThreadExecutor();
+        IAsyncExecutor asyncExecutor = new AsyncExecutor();
+        ICredentialsRepository credentialsRepository = new CredentialsLocalDataSource();
+        GetUserLastInsertedUsernameUseCase getUserLastInsertedUsernameUseCase =
```
ifoche: @manuelplazaspalacio I'm asking here the opinion from @xurxodev because I may be too tired but I wouldn't use a UseCase for this, but directly access the credentialsRepository with:
```
username.setText(credentialsRepository.getOrganisationCredentials().getUsername());
```
The dependency here is going towards "the kernel of the onion", so I guess it's not a problem...but @xurxodev may correct me saying that we need to use a UseCase for accessing the entities in any case...let's wait for his confirmation. In any case, does it make sense for you?

ifoche (changes requested):
@xurxodev can you please give a look to my comment in this PR? I think I have a CA doubt that you may easily clarify

app/src/ereferrals/java/org/eyeseetea/malariacare/strategies/LoginActivityStrategy.java
xurxodev: @manuelplazaspalacio @ifoche Yes we should to use a Use Case for retrieve data from database because it could entail additional application business rules that would be encapsulated in the use case. But a use case should not return primitive values, it is a code smell must return entities always. I have reviewed the code we already have in ereferrrals a use case that can serve you for what you need GetUserUserAccountUseCase that returns UserAccount and contains username

xurxodev (commented):
@manuelplazaspalacio @ifoche Yes we should to use a Use Case for retrieve data from database because it could entail additional application business rules that would be encapsulated in the use case. But a use case should not return primitive values, it is a code smell must return entities always. I have reviewed the code we already have in ereferrrals a use case that can serve you for what you need GetUserUserAccountUseCase that returns UserAccount and contains username

manuelplazaspalacio:
GetUserUserAccountUseCase don't return the last inserted credentials in eReferrals, changing String to credentials in the callback of the use case. @ifoche

ifoche (approved):
thanks @manuelplazaspalacio

PR #1659, opened 2017-11-13
Bug delay related hidden option

### LINKED ISSUE #1648: On RDT result question there is a huge delay between click and GUI reaction
Something very heavy is on the UI thread

manuelplazaspalacio:
Close #1648


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1663, opened 2017-11-13
Showing the auto-configure message only when the app is auto-configur…

### LINKED ISSUE #1647: Move auto-configuring message to actually when the app is auto-configuring and not by default
The goal is to avoid showing the message when we're not auto-configuring

manuelplazaspalacio:
Close #1647


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1669, opened 2017-11-15
When next field is dropdown when click on next, hide keyboard and cha…

### LINKED ISSUE #1621: When next field is a dropdown, hide keyboard and change focus
It currently skip dropdowns, but they shouldn't be skipped. Focus change to the dropdown and hide keyboard is the desired behavior

manuelplazaspalacio:
Close  #1621


ifoche (changes requested):
@manuelplazaspalacio at least in 7.1.1 I need to click twice for the dropdown to deploy. Can you take care of this? thanks!

manuelplazaspalacio:
Solved bug

ifoche (approved):
thanks @manuelplazaspalacio. Still think that we need to review a little bit the setupNextFocus and moveFocusToNext, but that can be done in a future refactor

PR #1670, opened 2017-11-15
Feature login types soft and full

Amejia481:
close #1655 
#  Full-login
## Demo Video
https://drive.google.com/file/d/1XL1shtYtggS5df5_wNy6h6o7JL7SIZoR/view?usp=sharing

### Without Clicking "Advanced Options" 
image

### Without Clicking "Advanced Options"
image

# Soft-login
## Demo video.
https://drive.google.com/open?id=1vcjSlkZYifg4FRfnrt4ATdozLb2UfbpM

### Without Clicking "Advanced Options" 

image

### Without Clicking "Advanced Options" 
image

PR #1672, opened 2017-11-19
Feature add db support for language translation

### LINKED ISSUE #1651: Dynamic strings
- [x] Add a DB table to manage strings & translations
- [x] Strings for options & questions will come from the local DB instead of Android default system
- [ ] Create CSV for strings and populate DB

Amejia481:
Close #1651


app/src/main/java/org/eyeseetea/malariacare/utils/Utils.java
```diff
@@ -58,44 +62,70 @@ public static String round(float base) {
...
+
+    @NonNull
+    public static String getInternationalizedString(@NonNull String key) {
+        Context ctx = PreferencesState.getInstance().getContext();
+        String language = getUserLanguage(ctx);
+        String translation = getLocalizedStringFromDB(key, language);
+
+        if (key == null) {
+            return key;
+        }
```
ifoche: @Amejia481 on null we were previously returning an empty string, and here you changed that behavior into return null. Is there any special reason for doing that? Did you check which consequences that could have? I would prefer to return anything but a null if possible, unless there's a good reason for not doing it

ifoche (changes requested):
@Amejia481 only a very minor comment. If you can check that returned null consequences that would be great. I will not merge this code until Login changes are applied to #1670 as this code comes from that branch. I have really appreciated the indexed solution for the strings. It's been really smart. Thank you for that.

app/src/main/java/org/eyeseetea/malariacare/utils/Utils.java
Amejia481: You're right, my bad, It looks like I missed this one.

Amejia481: @idelcano I updated the PR with the change, thanks for catching that one!

ifoche (approved):
thanks @Amejia481

PR #1673, opened 2017-11-21
Feature download languages from poeditor

### LINKED ISSUE #1652: POEditor API strings update
Given the POEditor endpoint for the strings, on app start, retrieve all the strings and update the strings & translations table

Amejia481:
Close #1652


ifoche (changes requested):
@Amejia481 there's a problem with this code. Each time there's a strings update, instead of overwriting the old ones, it creates new entries in the DB.
translations

The consequence is that select queries only find the first occurrence and therefore, strings never get changed. Can you please give a look to this?

Thanks!

Amejia481:
Sure @ifoche I on it!

Amejia481:
@ifoche I updated the PR please take a look!

ifoche (changes requested):
@Amejia481 looks like still happening. 

translations

Can you please take a look? Now the app looks like working, because when it searches for the translation it finds the last one that the user introduced in POEditor, but in reality if you look at the DB there are many translations there not being deleted.

Amejia481:
Did you remove the last database?

It's a little strange because when I simulate a string update, it deletes the whole language, before adding new strings:

For example: If you update the language **en**, the app will delete all the strings with the language code **en** before adding new strings.

Because PoEditor doesn't give you a specific string that was updated, it tells you the last time that a language was updated, and the app first deletes all the strings of this language and after add the new ones.


```
         for (Language language : toUpdate) {
              List<Term> terms = client.getTranslationBy(language.code);
  
            deletePreviousTranslations(language.code);
              for (Term term : terms) {
                 deletePreviousTranslations(language.code);
                  save(term, language.code);
              }
```
Maybe we can check it together to see what's the difference?

ifoche (approved):
thanks @Amejia481

ifoche:
BTW, just for the record, after your last change, that I asked you to review it again...I'm so sorry but I discovered that I was looking at the wrong DB. So thank you again for spending that time re-checking

PR #1678, opened 2017-11-28
Downloading metadata configurations from the server

### LINKED ISSUE #1653: Config file parser

Amejia481:
close #1653


app/src/androidTestereferrals/java/org/eyeseetea/malariacare/MetadataConfigurationDBImporterShould.java
```diff
@@ -0,0 +1,114 @@
...
+        } catch (Exception e) {
+            Assert.fail();
+        }
+
+
+        shouldBeInDB(16, 28, 28);
+    }
+
+    private void shouldNotBeAnyQuestionInTheDB() {
+        shouldBeInDB(0, 0, 0);
```
ifoche: @Amejia481 just like we did for the 1.0 version, this test is modifying the DB, and should use an in memory DB. Please give a look to PR #1757 and check out there how we finally made to use an alternative "staging" variant for the tests that uses an in memory DB. Can you bring that implementation to this PR and check that this test doesn't modify the app DB? Thanks!

app/src/ereferrals/java/org/eyeseetea/malariacare/data/sync/importer/metadata/configuration/IMetadataConfigurationDataSource.java
```diff
@@ -0,0 +1,11 @@
+package org.eyeseetea.malariacare.data.sync.importer.metadata.configuration;
+
+
+import org.eyeseetea.malariacare.domain.entity.Question;
+
+import java.util.List;
+
+public interface IMetadataConfigurationDataSource {
```
ifoche: If we want to follow the same arch that we used for the rest of the app, I think this interface should be in the main, and in the remote package. Then, from inside the variant we can implement that interface.

app/src/ereferrals/java/org/eyeseetea/malariacare/data/sync/importer/metadata/configuration/IMetadataConfigurationApi.java
```diff
@@ -0,0 +1,13 @@
+package org.eyeseetea.malariacare.data.sync.importer.metadata.configuration;
+
+
+import org.eyeseetea.malariacare.data.sync.importer.metadata.configuration.model.MetadataConfigurationsApi;
+
+import retrofit2.Call;
+import retrofit2.http.GET;
+
+public interface IMetadataConfigurationApi {
```
ifoche: To follow the naming convention I think we should rename this to xxxApiClientRetrofit just like with SurveyApiClientRetrofit

app/src/ereferrals/java/org/eyeseetea/malariacare/data/sync/importer/metadata/configuration/model/MetadataConfigurationsApi.java
```diff
@@ -0,0 +1,51 @@
+package org.eyeseetea.malariacare.data.sync.importer.metadata.configuration.model;
+
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class MetadataConfigurationsApi {
```
ifoche: Wouldn't it better to name this MetadataConfiguration? This is a model representing a MetadataConfiguration, and not an API, right?

app/src/main/java/org/eyeseetea/malariacare/data/database/converts/OptionConverterFromDomainModelToDB.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.eyeseetea.malariacare.data.database.model.OptionDB;
+import org.eyeseetea.malariacare.domain.boundary.converters.IConverter;
+import org.eyeseetea.malariacare.domain.entity.Option;
+
+public class OptionConverterFromDomainModelToDB implements IConverter<Option, OptionDB> {
+
+    @NotNull
+    @Override
+    public OptionDB convert(@NotNull Option domainModel) {
```
ifoche: Looks like you're following the visitor pattern here to make the conversion. If this is the case, we already have an standard naming convention for this that you can use. Check what's under sync -> exporter -> xxxVisitor. I think all that has to do with the IConverter could be adapted to the same naming convention

app/src/main/java/org/eyeseetea/malariacare/data/di/Injector.java
```diff
@@ -0,0 +1,100 @@
...
+import org.eyeseetea.malariacare.data.database.converts.OptionConverterFromDomainModelToDB;
+import org.eyeseetea.malariacare.data.database.converts.QuestionConverterFromDomainModelToDB;
+import org.eyeseetea.malariacare.domain.entity.Option;
+import org.eyeseetea.malariacare.domain.entity.Question;
+import org.eyeseetea.malariacare.utils.ConnectivityStatus;
+
+import okhttp3.OkHttpClient;
+import okhttp3.logging.HttpLoggingInterceptor;
+
+public class Injector {
```
ifoche: I'm a bit concerned about introducing this "Injector", that's even in a "di" package as if we were trying to use it as a dependency injector. I'm not sure here and I would like to know @xurxodev's opinion. Could we just create factories (that's something we've already used in the app before)? At least, this element is something that we should discuss with @xurxodev, that for sure.

app/src/main/java/org/eyeseetea/malariacare/data/sync/importer/poeditor/IPOEditorApi.java
```diff
@@ -0,0 +1,24 @@
...
+import org.eyeseetea.malariacare.data.sync.importer.poeditor.models.LanguagesResult;
+import org.eyeseetea.malariacare.data.sync.importer.poeditor.models.POEditorResponse;
+import org.eyeseetea.malariacare.data.sync.importer.poeditor.models.TermsResult;
+
+import retrofit2.Call;
+import retrofit2.http.Field;
+import retrofit2.http.FormUrlEncoded;
+import retrofit2.http.POST;
+
+public interface IPOEditorApi {
```
ifoche: Following the naming convention, should this be called xxxApiClientRetrofit?

app/src/main/java/org/eyeseetea/malariacare/domain/entity/Option.java
```diff
@@ -0,0 +1,58 @@
+package org.eyeseetea.malariacare.domain.entity;
+
+
+public class Option {
```
ifoche, line 4: This domain entity is anemic. This is something @xurxodev has already warned us about (http://xurxodev.com/clean-architecture-code-smells-parte-2/). Can you propose a way to provide this entity with some logic?

app/src/main/java/org/eyeseetea/malariacare/domain/entity/Question.java
```diff
@@ -0,0 +1,71 @@
+package org.eyeseetea.malariacare.domain.entity;
+
+
+import java.util.List;
+
+public class Question {
```
ifoche, line 6: Same comment as before. This entity looks anemic

ifoche (changes requested):
@Amejia481 this is just a first round of requested changes. I think with this PR we may even need some intervention of @xurxodev to agree with some architectural-dependent decisions. Can you please give a look to the comments? Thanks!

app/src/main/java/org/eyeseetea/malariacare/data/di/Injector.java
xurxodev: @ifoche I have looked this class a while, really this class is not a injector is a factory. I think we should rename this class and its package

app/src/androidTestereferrals/java/org/eyeseetea/malariacare/MetadataConfigurationDBImporterShould.java
Amejia481: Sure!

app/src/ereferrals/java/org/eyeseetea/malariacare/data/sync/importer/metadata/configuration/IMetadataConfigurationDataSource.java
Amejia481: Ok, I agree with moving the interface to main, but I am a little confuse with putting it on **data.remote** package because there's also a **data.sync.import** package where there's similar information.

For example the classes related to CVSImporter are in the same package **data.sync.import** but in the **ereferals** variant.

Could you help, please, to clarify in  which one should I put the interface?

app/src/ereferrals/java/org/eyeseetea/malariacare/data/sync/importer/metadata/configuration/model/MetadataConfigurationsApi.java
Amejia481: I named it API because it represents a model from the API response, but I can change it as you suggest

app/src/main/java/org/eyeseetea/malariacare/data/database/converts/OptionConverterFromDomainModelToDB.java
Amejia481: :+1:

Amejia481: @ifoche there also a similar name convention **xxxMaper** in the package *data.mappers* which one should I use?

ifoche: I've reviewed those mappers and I think your base implementation is closer to the visitors one, so I guess it would be much easier for everybody to understand it if you use the Visitor naming standard. However, maybe we can talk about which standard to use in our next meeting, so in the future, we're clear on which standard to follow

Amejia481: :+1:

ifoche (approved):
thanks @Amejia481

PR #1693, opened 2017-12-05
Feature stock survey screen

### LINKED ISSUE #1689: Create new stock surveys screen
This new tab, that will also appear based on a gradle variable, will list exactly with the same design as the normal sent/unsent lists, the stock surveys. For the stock surveys 2 things to take into account:
- Both sent/unsent will appear in the same list
- Surveys wont be accessible by clicking on them, they will just show all their DEs in te list preview

manuelplazaspalacio:
Close #1689


manuelplazaspalacio:
Close #1689

xurxodev (changes requested):
@manuelplazaspalacio I have found some things to review:
- New stock survey actions not working for me
- You are created conditions by isStockSurvey in DashboardAdapterStrategy but I think is better create a new separate adapter for stock and for this moment a strategy is not necessary. Is true that we should avoid duplicate code between DashboardAdapter and StockAdapter but we can use some approach to avoid it.
- the same for assessment_record I think that we should have separate layout for stock surveys in the list and reuse with include common layout parts

manuelplazaspalacio:
- New stock actions not working, this is made in the next issue.
- Removed changes from DashboardAdapterStrategy, DashboardFragment and assessment_record.
- Using common service to get SurveyDB, there are changes in next issues with repositories in next issues, I think is better to refactor this when we merge the next issues to not generate more problems.

manuelplazaspalacio:
For test it you can run this code in the main activitty:

`private void createStockSurveys() {
        for (int i = 0; i < 5; i++) {
            SurveyDB surveyDB = new SurveyDB(null, ProgramDB.getProgram("zaRb1RuqGVk"), null);
            surveyDB.save();
            new ValueDB("23", QuestionDB.findByUID("Y0rvgxJm28P"), surveyDB).save();
            new ValueDB("24", QuestionDB.findByUID("WA1ouIgQyCr"), surveyDB).save();
            new ValueDB("25", QuestionDB.findByUID("LLP04MYE6LM"), surveyDB).save();
            new ValueDB("26", QuestionDB.findByUID("wYekTzcJE6c"), surveyDB).save();
        }
    }`

xurxodev (approved):
thanks @manuelplazaspalacio. There is not an icon in stock survey list to identify the stock type but I hope this will be added in next issues, I will review it again.

Issue #1695, opened 2017-12-05
phone mask failing in ZW


ifoche:
with this example phone number: 0713336330 it should work and it doesn't

manuelplazaspalacio:
Solved in #1696

PR #1699, opened 2017-12-05
Solve bug when not answering a positive number question and is not ma…

### LINKED ISSUE #1696: non-compulsory questions avoid survey push
Survey cannot be sent if no value is provided for a non-compulsory question (like birthOrder in ZW program). If you click on the question, then you can because it sends an empty string instead of deleting the value

manuelplazaspalacio:
Close #1696

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/data/database/model/QuestionDB.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/data/database/model/QuestionDB.java
@@ -1152,6 +1152,9 @@ public class QuestionDB extends BaseModel {
 
     public void saveValuesText(String answer) {
         ValueDB valueDB = getValueBySession();
+        if (valueDB != null) {
+            valueDB.setValue(answer);
+        }
         SurveyDB surveyDB = (SurveyFragmentStrategy.getSessionSurveyByQuestion(this));
         SurveyFragmentStrategy.saveValuesText(valueDB, answer, this, surveyDB);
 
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/eyeseetea/malariacare/data/database/model/QuestionDB.java
```diff
@@ -1152,6 +1152,9 @@ public void saveValuesDDL(OptionDB optionDB, ValueDB valueDB) {
 
     public void saveValuesText(String answer) {
         ValueDB valueDB = getValueBySession();
+        if (valueDB != null) {
```
ifoche: Can you explain a bit more why this piece of code? It looks like we're doing twice the .setValue() call. I guess that it has to do with the issue #1695 that you say it's solved here? If that were the case, please remove this piece of code from this PR and create a separate PR for #1695, but anyway...please answer my question. Thanks!

ifoche (changes requested):
@manuelplazaspalacio I'm adding a "request changes" revision because I have a couple of doubts about this code:
1) It looks like there's an independent piece of code that solves #1695 but it's included here instead of creating a separate PR
2) There's a piece of code that I don't understand because we're setting the value twice, and I would like to understand better why that is necessary. Ideally, we shouldn't need to set a value twice

manuelplazaspalacio:
@ifoche 
1. I change this piece of code in this issue because when I was testing my changes, I put the question birthOrder to compulsory and I see It's not working well, and have to change that for make it work well. For me is related with this issue because is the way to make this view work well in all kind of cases.
2. Your are right we are setting the value twice , the error was in other place, because we are checking the value of the value db  instead of the text of the view when deleting the value when it's empty.

ifoche (changes requested):
@manuelplazaspalacio not sure if it's related or not with your changes, but if you click on birth order question and without introducing any data, you click on back button, the app crashes. Can you give a look to it?

ifoche (approved):
thanks @manuelplazaspalacio

PR #1700, opened 2017-12-05
solved bug surveys not scrolling in Android 4.4.2

### LINKED ISSUE #1697: Impossible to scroll on a 4.4.2 Android tablet

manuelplazaspalacio:
Close #1697

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/layout/listeners/SwipeTouchListener.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/layout/listeners/SwipeTouchListener.java
@@ -8,8 +8,8 @@ import android.view.MotionEvent;
 import android.view.View;
 import android.widget.ScrollView;
 
-import org.eyeseetea.malariacare.R;
-import org.eyeseetea.malariacare.data.database.utils.PreferencesState;
+import org.eyeseetea.malariacare.layout.listeners.strategies.ASwipeTouchListenerStrategy;
+import org.eyeseetea.malariacare.layout.listeners.strategies.SwipeTouchListenerStrategy;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/eyeseetea/malariacare/layout/listeners/SwipeTouchListener.java
```diff
@@ -43,7 +43,7 @@ public boolean onTouch(View v, MotionEvent event) {
             fixScrollEventY(event);
             scrollView.onTouchEvent(event);
         }
-        return gestureDetector.onTouchEvent(event);
+        return false;
```
ifoche: @manuelplazaspalacio I have a doubt here. This Listener is in the main and affects also the rest of the variants. returning false wouldn't it make the rest of the variants to stop working? The feature I'm concerned about is the previous/next swipe in Laos, for example. I guess that we would need to create a Strategy here, moving the common behavior to a common method and separating the difference into a per-variant answer?

ifoche (changes requested):
@manuelplazaspalacio just a little doubt about the possible need of a strategy in the swipe gesture detector change that you implemented.

app/src/main/java/org/eyeseetea/malariacare/layout/listeners/SwipeTouchListener.java
manuelplazaspalacio: @ifoche Finally I add the strategy to not change the behaviour in other variants. I am checking and returning false on onDown method in the swipeTouchListener solve the problem, I don't understand why we  have to consume the motion event. I think we have to review the SwipeTouchListener code for other variants with Android version 4.4.2, because I'm not sure that scroll is working correctly.

manuelplazaspalacio:
@ifoche Creating strategy , but i think we have to review the swipe touch listener in other variants, probably the scroll it's not working well in 4.4 Android version. I think we don't have to consume the touch event in all cases  we are consuming it.

ifoche (approved):
thanks @manuelplazaspalacio

PR #1701, opened 2017-12-06
Not putting surveys as completed if there is an error when pushing su…

### LINKED ISSUE #1698: Duplicates error
Despite we set the set the web service to always reply with success if there is an error in the process of sending data to the web service (for example the web service did not responded, or the connection was interrupted), the android app will enter in a loop where all the pending records are sent again and again.

manuelplazaspalacio:
Close #1698


manuelplazaspalacio:
@ifoche  checking network errors from retrofit.

ifoche (approved):
thanks @manuelplazaspalacio

PR #1702, opened 2017-12-07
Feature config file downloader

### LINKED ISSUE #1654: Config file downloader

Amejia481:
Closes #1654


ifoche (commented):
As there are some dependencies with #1653 issue, still not asking for changes here, but in a first try, I've seen problems with "yes only" questions and proper recognition of the country associated to the user

ifoche (approved):
thanks @Amejia481. I've made some changes and other should be done, but to avoid future big merges I prefer to accept the code as-is now, and ask you those changes separately in the issues #1763, #1764, #1765, #1766. Can you please take care of them? Thanks!

PR #1707, opened 2017-12-13
Feature expense stock normal survey

### LINKED ISSUE #1694: Automatically create expense stock survey with normal MCS survey

manuelplazaspalacio:
Close #1694


xurxodev (changes requested):
@manuelplazaspalacio I have found two problems:
- If when a new survey is created then a stock survey is created, when a survey is deleted then the related stock survey should be removed
- autogenerated stock survey show 3 fields in list and manual stock surveys show 4 fields in list

manuelplazaspalacio:
Not showing ASMQ 25/50 because this question is not  in malaria survey at this moment is added in issue #1712. Now only showing questions with relation in malariasurvey, @ifoche  created an issue to deal with it #1860.

xurxodev (changes requested):
@manuelplazaspalacio We have two ways to remove a survey:
- When you press back before to complete a survey, this is ok. Now when press back in this case stock survey is removed also.
- When a survey is completed and appears on the unsent list, is possible swipe left to remove it, is this case stock survey is not removed

xurxodev (approved):
thanks @manuelplazaspalacio

PR #1708, opened 2017-12-14
Feature modify pull push stock surveys

### LINKED ISSUE #1691: Modify pull/push/dbflow queries/visitors to contemplate new program/surveys
The main goal of this task is to be able to launch the visitors to translate stock surveys into dhis DB and manage these surveys in queue just like the rest of surveys. 

The main difference with stock surveys is that they do not contain control DEs

The reason why we need to check the rest of elements is that they may be prepared in their queries for having only one program in the DBs, so we need to be sure that we're retrieving on each query the appropriate questions (only for the program that we want).

manuelplazaspalacio:
Close #1691


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1728, opened 2018-01-08
Feature add config version payload

### LINKED ISSUE #1719: Send configVersion in the push payload
When sending a voucher, we need to add a new "configVersion" key in the json, indicating the version number of the config file that's stored in the settings, and checked on each app start.

manuelplazaspalacio:
Close #1719


ifoche (approved):
thanks @manuelplazaspalacio

PR #1734, opened 2018-01-10
Putting to conflict status surveys when server returns failed or not …

### LINKED ISSUE #1732: Treat individually each survey error on push answer for multiple surveys
Move to corresponding state per-survey instead of all the surveys as a whole

manuelplazaspalacio:
Close #1732


app/src/androidTestEreferrals/java/org/eyeseetea/malariacare/data/sync/exporter/WSPushControllerShould.java
```diff
@@ -0,0 +1,112 @@
...
+        for (String eventUID : eventUIDs) {
+            SurveyDB surveyDB = new SurveyDB(new OrgUnitDB("test"), new ProgramDB("test"),
+                    new UserDB("test", "test"));
+            surveyDB.setStatus(Constants.SURVEY_COMPLETED);
+            surveyDB.setEventUid(eventUID);
+            surveyDB.save();
+            surveysIDs.add(surveyDB.getId_survey());
+        }
+    }
+
```
ifoche, line 166: @manuelplazaspalacio if you execute the test and then try to open the app, the app will crash. Ideally the test should work completely in memory, and not using the app's DB, otherwise subsequent tests will depend on the exit state of the previous ones, or in the worst scenarios, things like this can happen. Can you please change it either to implement the use of an in-memory DB for the tests (the best) either to reset the state of the DB at the end of the test? If you go for the 2nd option because of the lack of time, please remember to create an issue to refactor that in the future

ifoche (changes requested):
@manuelplazaspalacio please take a look to the inline comment. Thanks!

ifoche (changes requested):
Preferences of the app are modified during the test but not restored after the test.

ifoche (approved):
thanks @manuelplazaspalacio and @idelcano

PR #1737, opened 2018-01-10
Deleting sent surveys older than 2 weeks #1730

### LINKED ISSUE #1730: Maintenance deletion
Automatically delete any sent event older than 2 weeks on app start

Amejia481:
close #1730


app/build.gradle
```diff
@@ -47,9 +47,12 @@ android {
         release {
             minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+            buildConfigField "int", "MaxDayBeforeDeletingSentSurvey", "14"
+
```
ifoche: @Amejia481 this applies to all variants, so if we accept this code, all variants would suddenly start deleting sent surveys older than 14. I think the beauty of including the gradle variable would be to do it only for target variant (ereferrals). Maybe you can change it into a boolean var deciding if the variant must performMaintenanceTasks and apply the surveys removal only if that variable is true

app/src/main/java/org/eyeseetea/malariacare/SplashScreenActivity.java
```diff
@@ -67,6 +68,9 @@ private void init() {
             splashActivityStrategy.initPullFilters(pullFilters);
             splashActivityStrategy.executePull(pullUseCase, pullFilters);
         }
+
+        SurveyDB.deleteOlderSentSurveys();
```
ifoche: As explained before this is in the main, so I think we would need to either change the gradle var to the convenient variant, or create a strategy to create a performMaintenanceTasks method that could be altered depending on the variant.

app/src/main/java/org/eyeseetea/malariacare/data/database/model/SurveyDB.java
```diff
@@ -1055,6 +1060,37 @@ public static void deleteAll() {
...
+        List<SurveyDB> sentSurveys = getAllSentSurveys();
+
+        for (SurveyDB surveyDB : sentSurveys) {
+
+            if (isSurveyOlderThan(surveyDB, MaxDaysBeforeDeletingSentSurveys)) {
+
+                new Delete().from(ValueDB.class).where(
+                        ValueDB_Table.id_survey_fk.eq(surveyDB.getId_survey()));
+
+                surveyDB.delete();
```
ifoche: @Amejia481 please instead of getting all surveys and looping on all of then manually matching those that are older than the given date, please create directly a DBFlow call returning only those surveys older than a given time. Otherwise we would be introducing a performance problem.

ifoche (changes requested):
@Amejia481 I've asked for some changes in the code, please give a look at the comments and let me know if you need any clarification. Apart from the code changes, I've done a functional test and I'm not able to make it work. Please let me describe what I've done just in case I'm missing something.

1) I've created a survey
screenshot_20180125-011643

2) I've manually changed the device date more than 2 weeks in the future from the survey date
screenshot_20180128-011931

3) I kill the app and open it again to force it passing through the splash screen

4) If I'm not bad I shouldn't see the survey in the listing, as it should be deleted. But I can still see it there...am I doing something wrong?

app/build.gradle
Amejia481: You are correct, I made the change.

app/src/main/java/org/eyeseetea/malariacare/data/database/model/SurveyDB.java
Amejia481: I was trying to find a way of applying this  sql statement with dbflow 
~~~~
Where event_date < date('now','-14 years')
~~~~

But I was unable to do the same with DBFlow, I will research again.

Amejia481:
I added all the improvements suggested and I simulated the functional test as you did.
It is working fine :grinning: !

**1.I've created a survey**
image

**2.I've manually changed the device date more than 2 weeks in the future from the survey date**
image

**3.I kill the app and open it again to force it passing through the splash screen**
image

image

ifoche (approved):
thanks @Amejia481. Works like a charm. Only some minor cosmetic changes added

Amejia481:
:+1:

PR #1754, opened 2018-01-14
Fixing issue #1751

### LINKED ISSUE #1751: Exit from a sent survey takes you to the 2nd tab instead of the first one

Amejia481:
close #1751

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -98,6 +98,7 @@ android {
             buildConfigField "String", "defaultLocale", "\"en\""
             buildConfigField "int", "maxDaysForDeletingSentSurveys", "14"
             buildConfigField "boolean", "performMaintenanceTasks", "false"
+            buildConfigField "boolean", "goBackOnASentSurveyTakesYouTo2ndScreen", "true"
         }
         laos {
             applicationId 'org.eyeseetea.surveillance_la_bb'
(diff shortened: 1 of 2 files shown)


Amejia481:
Hi @idelcano!

I saw that you added this behavior to the [this commit ](https://github.com/EyeSeeTea/pictureapp/commit/040583d847771e83cd2522486a0ee4b0d43265ff), please, could you confirm us in case this behavior was expected in another variant of the app?

This was the specific snippets of code that cause this behavior.

 **tabHost.setCurrentTabByTag(getResources().getString(R.string.tab_tag_improve));**

```
     if(isSent){
          tabHost.setCurrentTabByTag(getResources().getString(R.string.tab_tag_improve));
          initAssess();
          initImprove();
      }
      else{
          initAssess();
          unsentFragment.reloadData();
      }
```

idelcano:
what is the problem with this code?
i think it's the expected behavior because it's in the main... I think it
is correct for apps with local sent survey list (shown in improve tab)

Amejia481:
@ifoche  Reported this issue "Exit from a sent survey takes you to the 2nd tab instead of the first one #1751" for this reason I asked

idelcano:
@Amejia481 

I think this is a expected behaviour for other variants. But i don't remember very well..

For example for this case:

If you open a not sent survey, and this survey changes to sent when you are reading it...

That survey will never be shown in assets list anymore, at least in other variants...
In that case to see the survey after close the surveyfragment, you need move to the improve tab.
If you don't move to the improve tab, the user might think that the survey disappeared.

ifoche (approved):
thanks @Amejia481 . I kept the showUnsentFragment call as it was because there was an issue with headers reload in other variant (myanmar if I'm not bad) that forced us to make it this way

Amejia481:
:+1:

PR #1756, opened 2018-01-15
Showing error messages when there is an error downloading the media f…

### LINKED ISSUE #1753: Library doesn't show the message "donwloading"

manuelplazaspalacio:
Close #1753


ifoche (changes requested):
@manuelplazaspalacio message is not updated when wifi connexion is activated. Can you give a look at it? Thanks!

ifoche (approved):
thanks @manuelplazaspalacio

PR #1758, opened 2018-01-21
Feature implement question threshold on dynamic tab adapter

### LINKED ISSUE #1727: Implement QuestionThreshold support on DynamicTabAdapter

Amejia481:
close #1727


Amejia481:
I found a bug in the survey screen on the threshold part:

When you have a question with two parents one a **QuestionOption** and another a **QuestionThreshold** and the **QuestionOption** makes the question visible but the **QuestionThreshold** hides it because its condition has not been fulfilled, and you are unable to submit the survey.

When you press the *Next* button the app shows a message that reads **you must complete all the required fields**, because only one of the two  conditions are fulfilled that of the  **QuestionOption**

xurxodev (approved):
thanks @Amejia481

PR #1760, opened 2018-01-23
Allowing to access with demo credentials not touching the demo button.

### LINKED ISSUE #1759: On demo program, credentials doesn't allow you to access
To replicate this bug, enter demo mode, then kill the app and open it again. You'll be asked for credentials. I think we should bypass the login when we're in demo mode

manuelplazaspalacio:
Close #1759


ifoche (changes requested):
@manuelplazaspalacio it's still asking me for credentials with demo user. Just to be sure, what I'm doing is to:
1) access demo mode
2) kill the app
3) open the app again. Instead of the soft login screen I should be watching the dashboard screen

Can you please look at it?

manuelplazaspalacio:
Changes done

ifoche (approved):
thanks @manuelplazaspalacio

PR #1762, opened 2018-01-24
Not allowing to enter in app if android version lower than 5.0.

### LINKED ISSUE #1761: If Android version < 5.0 show pop-up and block app access
On app start, if the Android version is less than 5.0, we must show a pop-up saying "Apologizes but Connect is not compatible with your device Android version. You must have Android 5.0+ to get Connect working.". Only ok must be shown as option, and after clicking on ok, the app must close.

manuelplazaspalacio:
Close #1761


ifoche (approved):
thanks @manuelplazaspalacio

PR #1780, opened 2018-01-25
Bringing the behavior from previous commit issue #1763

### LINKED ISSUE #1763: Review program metadata update process
I've been playing with Postman and the metadata config file in the dataStore. I can only get the config file downloaded and parsed the first time. After that, if I change the version both in the countries_versions config file and in the config file for a given country, kill the app and enter again, I don't get the config file updated. I've tested with T_TZ, changing the default config in clone server (to avoid the collision between T_TZ and TZ) and using 8001 user.

Amejia481:
close #1763


ifoche (approved):
thanks @Amejia481

PR #1786, opened 2018-01-25
Adding default language to translate if there is not translation in t…

### LINKED ISSUE #1776: Set default language
Get from system if possible, otherwise assume English

manuelplazaspalacio:
Close #1776


ifoche (approved):
thanks @manuelplazaspalacio

PR #1788, opened 2018-01-25
Showing detailed message on POEditorError

### LINKED ISSUE #1770: Manage errors on POEditor tokens failing

manuelplazaspalacio:
Close #1770


xurxodev (changes requested):
@manuelplazaspalacio I have found some changes to realize:
- Domain exceptions should not have technology names because should not exists coupled, please rename POEditorException
- Review HTTP codes that you are using to throw exception, you are using 4011
- Splash screen activity should catch this new exception when downloading languages

manuelplazaspalacio:
Changes done, error code 4011 is the returned when POEditorToken is wrong.

xurxodev (approved):
thanks @manuelplazaspalacio

PR #1807, opened 2018-02-05
Add queueDisplayPriority to config files #1805

### LINKED ISSUE #1805: Add queueDisplayPriority to config files
In order to determine which questions are VISIBLE, INVISIBLE and IMPORTANT a new tag called "queueDisplayPriority" need to be added. In case no tag is present, VISIBLE will be assumed. In case no IMPORTANT is present, we will continue working as we currently do

Amejia481:
close #1805


ifoche (approved):
thanks @Amejia481

PR #1816, opened 2018-02-08
Change login behavior

### LINKED ISSUE #1810: Change login behavior
Sometimes config file is properly downloaded but strings aren't, or vice-versa. To stop with strange behaviors the following changes are required:

1) Translations tables shouldn't be deleted until new ones are already downloaded and parsed from POEditor. The alpha testers has reported situations (with variable internet connexion) where he had the strings properly downloaded, and after login again he didn't. So he probably got his Strings deleted while new ones where not yet properly downloaded. (I have a DB from alpha tester, so please ask me for it).

2) On full login: User must only proceed to dashboard if both config and strings are properly downloaded. Otherwise, we must show a Toast with the error and get the user back to login.

3) On soft login. If config file download fails, read-only mode as we're doing now. If translations download fails, a pop-up saying "Warning, strings download failed" should show.

Amejia481:
close  #1810


ifoche (approved):
thanks @Amejia481

PR #1818, opened 2018-02-09
Adding language to URL on webviews #1813

### LINKED ISSUE #1813: Add language to URL on webviews
Please add the parameter...
```
&lang=<code>

where <code> is es, en, pt...
```
...to the end of the url loaded into the webviews. Please make sure to pass the language code that is defined in settings or the default language in case system default option is configured

Amejia481:
close #1813


ifoche (approved):
thanks @Amejia481

PR #1823, opened 2018-02-13
Only downloading metadata when is not demo #1821

### LINKED ISSUE #1821: demo only works when you enter previously a normal user
If you install from zero the app and directly enter demo it seems to stay forever and crash after a while when you try to create a survey

Amejia481:
close #1821


ifoche (approved):
thanks @Amejia481

PR #1824, opened 2018-02-14
Showing error message when unable to open a unsupported file

### LINKED ISSUE #1822: Avoid opening unsupported extensions
unsupported formats like PDF shouldn't try to open. Add a minimum validation before launching the intent based on the extension of the file

Amejia481:
close #1822


ifoche (changes requested):
@Amejia481 there's a new requirement here. Apparently, the apps crashing on the grid view when trying to preview anything that's not an image or a video. Could you please add a "No preview available" free image for these cases? Thanks!

Amejia481:
Sure, I'll do it!

Amejia481:
Done!
screenshot_1518621626

ifoche (approved):
thanks @Amejia481. I've noticed that anyone who may have a pdf previously downloaded could still have a crash because the media was downloaded and not identified in the DB as unknown. However I'm not asking for any migration here because I think there's no pdf currently on the field, so I will simply prevent the client. Thank you for this code.

PR #1827, opened 2018-02-14
Fixing Simple/Advance toggle on press clear & logout

### LINKED ISSUE #1826: During a soft login, clear & logout doesn't toggle simple/advance switch
If during a soft login you click on clear & logout you're taken to the full login screen collapsed (so simple) but the simple/advance switch says "Simple", so we're in a Simple/Advance switch inversion situation

Amejia481:
close #1826


ifoche (approved):
thanks @Amejia481

PR #1828, opened 2018-02-14
Not sending the user to the login activity after push conflict

### LINKED ISSUE #1825: A conflict sending takes the user to login page
A conflict sends the user to login without leaving him the time to read the message saying that a conflict occurred. In principle, the user shouldn't be taken to login screen in case of conflict.

Amejia481:
close #1825


ifoche (approved):
thanks @Amejia481

PR #1840, opened 2018-02-21
Feature download ou group associated to assigned ou

### LINKED ISSUE #1838: Download OU Groups associated to the assigned OU
When the OU is selected by the user or automatically selected by the system, we need to download the OU Group(s) that OU belongs to and save it to a preference in order to later decide which program to use when creating a survey.

- [x] Download when automatically OU is selected by the system
- [x] Download when the OU is selected by the user

Amejia481:
close #1838


Amejia481:
To see if the ou group are been downloaded and save go to the preference file `org.eyeseetea.surveillance_cnm_bb_preferences.xml`

They keys are: 
-  org_unit_group_id
-  org_unit_group_code

ifoche (approved):
thanks @Amejia481

PR #1853, opened 2018-03-07
Put surveys on SENT state after a configFileObsolete exception

### LINKED ISSUE #1852: Problems on 209 code treatment
- [x] The survey is treated as conflict (it should be sent)
- [x] After a while (changing tabs for example) it allows the user to create surveys again (exits the read-only mode)
- [x] It doesn't force the survey upgrade automatically

ifoche:
closes #1852


xurxodev (approved):
thanks @manuelplazaspalacio and @ifoche

PR #1879, opened 2018-03-12
Showing * if stock surveys unsent.

### LINKED ISSUE #1872: We should differentiate send and unsent stock surveys in list

manuelplazaspalacio:
Close #1872


ifoche (changes requested):
@manuelplazaspalacio surveys are not being refreshed on push. I think we would probably need a StockSurveyService, just like with SurveyService, and call to reload() on that fragment when a push is completed. Can you take care of this? Thanks!

ifoche (approved):
thanks @manuelplazaspalacio

PR #1926, opened 2018-03-13
Solved bug with different time in normal survey and stock survey.

### LINKED ISSUE #1906: Time for normal survey and stock one not necesarily the same

manuelplazaspalacio:
Close #1906


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1927, opened 2018-03-13
Adding ou and program to action bar

### LINKED ISSUE #1909: actionbar should show selected OU and program

Amejia481:
close #1909


xurxodev (changes requested):
@Amejia481 we have to add two concepts to toolbar:
- The selected org unit and this is ok
- The selected program, we should add the program code also to the toolbar, for example:
Org unit name - program code

Amejia481:
Hi @xurxodev, 

**The orgUnit name is too big to fit on the same line with the program, as you can see in the image below:**

image

**Can I add it in two lines, OrgUnit name on the first line and the second the program?**

image

xurxodev:
@Amejia481 d7iipQZaHad is not the code for the program, It's a uid, the code is HC that It's smaller, we should not remove app name. I think is better set a smaller font to orgunit name - program code if is necessary.

Amejia481:
Got it!

Amejia481:
Thanks @xurxodev, now it fits:

image

Amejia481:
@xurxodev it's done!

xurxodev (approved):
thanks @Amejia481. I have refactored your code a little to use baseActivityStrategy because this customization on the toolbar (org unit - program) is only for CNM

Amejia481:
Thanks @xurxodev !

Issue #1932, opened 2018-03-13
when try to delete vmw the app crash


idelcano:
steps:
login with ifoche_cambodia.
Select Banteay Neang org unit.
create new survey.
delete survey.

PR #1933, opened 2018-03-13
fix crash when try to delete null stock survey

idelcano:
Closes https://github.com/EyeSeeTea/pictureapp/issues/1932


xurxodev (approved):
thanks @idelcano

PR #1936, opened 2018-03-14
Increasing stock fab

### LINKED ISSUE #1912: Stock + button should be higher than aux buttons

Amejia481:
close #1912


xurxodev (approved):
thanks @Amejia481

PR #1938, opened 2018-03-14
Feature change about and copyright

### LINKED ISSUE #1928: Modify copyright & about us text

manuelplazaspalacio:
Close #1928


ifoche (changes requested):
@manuelplazaspalacio looks like it's showing the all about yet, right?
screenshot_20180314-110140

ifoche:
I think it's because there's a French version in the main and it takes that when it doesn't find the variant one. I guess that either we pass the French version from the main to all the variants but this one, either we create a French version in CNM, but it doesn't make sense to have a wrong French version. Thanks!!

ifoche (changes requested):
@manuelplazaspalacio looks like the text is not exactly the same as in the document. At least EST appears twice. Can you check it please?

screenshot_20180314-135227

ifoche (approved):
thanks @manuelplazaspalacio

PR #1947, opened 2018-03-14
when user ou is null use the first program

### LINKED ISSUE #1920: Create new survey when no OU is assigned produces a crash

idelcano:
Closes #1920 
I used try catch (IllegalArgumentException e)  to avoid modify common code  with other variants


xurxodev (approved):
thanks @idelcano

PR #1950, opened 2018-03-14
Adding Quarantine surveys to note sent stock surveys adapter.

### LINKED ISSUE #1942: Asterisk not showing in not sent states for stock surveys
Now that server is down, it happened to me that I after creating some stock surveys, changing to another tab and back again to stock tab showed a stock survey without the asterisk. In addition, I could delete the stock survey, so it was clear that survey was not in sent state. I wonder if we are showing the asterisk for all states different than sent, or we're missing a state

manuelplazaspalacio:
Close #1942


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1951, opened 2018-03-14
Center cnm questions

### LINKED ISSUE #1914: Center images from options where left alignment is not what's in the design

idelcano:
Closes #1914


xurxodev (approved):
thanks @idelcano and @Amejia481

PR #1954, opened 2018-03-14
Use a new SaveValueUseCase for each saved value to avoid duplicate va…

### LINKED ISSUE #1952: duplicated foreign keys on some stock surveys

idelcano:
Closes #1952 

the problem was in the SaveValueUseCase. 
The expected workflow is:

SaveValueuseCase.execute(value1);
SaveValueUseCase.run(); ->value1 saved
SaveValueuseCase.execute(value2);
SaveValueUseCase.run(); ->value2 saved

But in some cases the execute is called by multiple listeners at same time, and the log show this workflow:

SaveValueuseCase.execute(value1);
SaveValueuseCase.execute(value2);
SaveValueUseCase.run(); ->value2 saved
SaveValueUseCase.run(); ->value2 saved

I created a SaveValueUseCase for each value and it works now.

@manuelplazaspalacio  thanks for the help.


xurxodev (approved):
thanks @idelcano. 
I have not reproduced the duplicates in development. I have not seen any problems at the operating level with this solution but at the architecture level. I will create an issue to refactor at the architecture level.

PR #1956, opened 2018-03-15
Feature bring quarantine workflow and services changes from development

### LINKED ISSUE #1874: Bring quarantine workflow and services changes from development to SCMM 1.0
There are changes in the quarantine treatment process that were done during the last year of development that need to be introduced in the old v1.0_mm branch. Concretely some of them can be seen in the following files:
- PushService.java
- AlarmPushReceiver.java
- SurveyChecker.java

The goal of this task is to get rid of 2 different issues that users have reported during the last year:
- Surveys that are sent but app interface keep the asterisk as not sent
- Surveys that stay in the app and are never sent while in the app look like sent

This is the main task of this release, so please take your time to test it carefully and spend several time generating corner cases (network problems, server down, user doing crazy interactions...)

idelcano:
Closes #1874


xurxodev (approved):
thanks @idelcano

PR #1967, opened 2018-03-20
Solving bug when selecting two times the rdt invalid and putting thre…

### LINKED ISSUE #1876: Stock calculation error when counters are involved
To replicate the bug just do the following on a clean installation:
1. Create an invalid test (so 2 invalid tests)
2. Look at the monitoring table. You'll see 3 RDTs used, but on 2 tests, we spend only 2 RDTs. 

Looking at the DB, the value is there, so it looks like a calculation problem and not a presentation one. Probably we're incrementing 1 during the last confirmation of invalid and we shouldn't. 

Some testing around the different invalid situations must be done to be sure that we don't introduce a new bug while fixing this one.

manuelplazaspalacio:
Close #1876


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1972, opened 2018-03-21
Migration for treatment table.

### LINKED ISSUE #1971: Treatment matrix migration
In the best case, we should delete treatment-related tables, create and populate. If best case is not possible then delete the database and create again.

manuelplazaspalacio:
Close #1971


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1997, opened 2018-03-28
Feature fix push answerratio bug

### LINKED ISSUE #1991: Remove UnkUvjuDrVO DE from program stage at dev server and test the app behavior

idelcano:
Closes #1991 
The answer ratio is wrong, and the surveys without treatment can't be completed.  I fixed it setting the start required question count from 1 to 0 for multiquestiontab because the first question added is "parent name"....

I fixed the starter required count and all the surveys can be completed now. But when you have 8 required values the answered compulsory number is 8-11, but the answered compulsory count should be as max the number of required questions(8). (i talked it with @ifoche)

I fixed a repeated call to updateSurveyStatus and a wrong order to set the stock survey as completed(first needs fill the survey, and after complete it).


ifoche (approved):
thanks @idelcano

PR #2009, opened 2018-03-30
Substitute CSV Version by Config Version in about text

### LINKED ISSUE #2008: Substitute CSV Version by Config Version in about text
We currently have a CSV Version in the app that is no longer in use. Instead of that, we would need to change that by "Config Version: " and use the number that we have stored in preferences indicating the version number for the selected program.

Amejia481:
close #2008


ifoche (approved):
thanks @Amejia481

PR #2018, opened 2018-04-02
Getting orgunits assigned to the logged user, and deleting the previo…

### LINKED ISSUE #2015: If you assign a village OU to a user, it's not proposed in the auto-configuration of the OU after login. Only HC OUs appear

manuelplazaspalacio:
Close #2015


xurxodev (changes requested):
@manuelplazaspalacio this solution work if you are autoconfigured and do log in from settings but if autoconfiguration fail because your IMEI does not exists in any org unit then when do log in always fail showing invalid credentials, the problem is logout without previous autoconfiguration

xurxodev (approved):
thanks @manuelplazaspalacio

PR #2036, opened 2018-04-16
Removing log interceptor in eReferralsAPIClient to solve never ending…

### LINKED ISSUE #2032: Connect sending forever
The app is showing the send progress wheel forever and never sends

manuelplazaspalacio:
Close #2032


xurxodev (changes requested):
@manuelplazaspalacio Ok It's true that work but We need interceptor for debug purposes. So can you investigate how we can add it again without push is failing?:
- Seems that bug is in retrofit with the interceptor. We have changed library versions to integrate bugshaker, is possible that library versions are the problem? 
Please, take some time to investigate it? Thanks

xurxodev (approved):
thanks @idelcano

Issue #2035, opened 2018-04-16
Introduce new CNM stock control tab


ifoche:
I've just shared the design with you. The idea is that this module will be a new one with the following properties:
- It will substitute the current stock module (don't pay attention to the fact that in the design there are 4 tabs, there will only be 3)
- It will contain the + button that is currently in the stock module (with all its functionality)
- It must be in the middle of a monitoring tab and a listview tab. It's better if we can implement this in a pure Android way (no HTML with all its complications), but we need to provide the same look&feel that the monitoring tables are already providing (but with the new design).

PR #2037, opened 2018-04-17
Feature cnm stock control screen

manuelplazaspalacio:
Connect #2035


manuelplazaspalacio:
Close #2035

xurxodev (changes requested):
@manuelplazaspalacio stock table is empty because an exception occurs in the async thread, can you review it?

xurxodev (approved):
thanks @manuelplazaspalacio. I have fixed a bug and rename some classes to more semantic name

PR #2049, opened 2018-05-01
Feature add telephone number

### LINKED ISSUE #2044: Add telephone number for both VMW & HC
Please introduce as the last question about the patient information, just before the treatment info.

Please don't forget the DB migration (rude one, deleting and creating everthing from zero...we can reuse the existing one)

idelcano:
Closes #2044


xurxodev (approved):
thanks @idelcano. I have some doubts, but I will wait for the testing to comment

PR #2084, opened 2018-05-07
Solved bug not going to review screen when coming from review screen.

### LINKED ISSUE #2064: In read-only mode after entering in any question, you can never go back to review screen

manuelplazaspalacio:
Close #2064


xurxodev (approved):
thanks @manuelplazaspalacio

PR #2085, opened 2018-05-07
Solved bug when going to a question from review screen.

### LINKED ISSUE #2073: Crash navigating to a question in read-only mode
Maybe related to the new question, but doesn't look like

manuelplazaspalacio:
Close #2073


xurxodev (approved):
thanks @manuelplazaspalacio

PR #2090, opened 2018-05-07
run autoconfig only if ou is not selected

### LINKED ISSUE #2066: Auto-configure is always happening
On boot, when auto-configuring, only do it if there's not an org unit already configured

idelcano:
Closes #2066


ifoche:
I'm no longer able to replicate the issue that originated this issue. Closing the issue

PR #2103, opened 2018-05-14
Opening uncompleted survey when putting app to in background.

### LINKED ISSUE #2099: Save form when app focus changes
We're currently saving the form we're filling in when the screen switches off, so if the user turns it on again, the form is not lost. However, we're not doing it when the user changes to another app or press the Android home button.

manuelplazaspalacio:
Close #2099


xurxodev (approved):
thanks @manuelplazaspalacio

PR #2104, opened 2018-05-14
fix empty settings value

### LINKED ISSUE #2096: Fonts customize doesn't persist

idelcano:
Closes #2096


xurxodev (commented):
@idelcano I don't understand this solution according to issue description. I have tested issue problem in release branch and I have not reproduced. do you have reproduced the error? can you explain me this solution?

idelcano:
@xurxodev yes I can replicate.
The problem is with the font preference in settings activity is always
empty when you open the settings activity.

I needed use the setsummary to make visible the preference name.

idelcano:
antesydespues

xurxodev (approved):
ok It's true, the summary no persist I understood that custom font functionality does not persist but this work sucessfully after close and re-open the app. 
Thanks @idelcano

PR #2105, opened 2018-05-14
Added custom dropdown cosmetic in connect

### LINKED ISSUE #2097: Improve design of dropdowns
- [x] Make padding larger
- [x] Add separator
- [x] Identify (either with a background color or another method) the selected option

When possible, it would be even better if we can just make it as closest as possible to [this example](https://camo.githubusercontent.com/2a00581fffe96cff4e3c33896c9de4a811eb29d0/68747470733a2f2f776166666c65696f2d6469726563742d75706c6f6164732d70726f64756374696f6e2e73332e616d617a6f6e6177732e636f6d2f75706c6f6164732f3539616436626464366437366264303031323832386466622f313235353136633636653832633732386163653231653064343664623937383832363837386462613837653661623763663630646131636337393132373632323634353566373762323364366233366166333030366231333030303531626666303534613535353666396630386232396632653736613731643830393033653564383365356637346335333431306664656564363164626631336136333263393165646265326533643731353861626437663337656439393865343933343166303031613861666235393364613439663432633035316632353961632e706e67)

idelcano:
Closes #2097

### DIFF
--- a/app/src/cambodia/java/org/eyeseetea/malariacare/strategies/DropdownAdapterStrategy.java
+++ b/app/src/cambodia/java/org/eyeseetea/malariacare/strategies/DropdownAdapterStrategy.java
@@ -1,0 +1,12 @@
+package org.eyeseetea.malariacare.strategies;
+
+import android.view.View;
+import android.widget.ListView;
+
+import org.eyeseetea.malariacare.R;
+
+public class DropdownAdapterStrategy {
+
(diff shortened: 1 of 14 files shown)


xurxodev (changes requested):
@idelcano On this case I think that if we use styles and variants is a better option that uses strategies because Java code is not necessary.
Please review this link https://stackoverflow.com/questions/4569751/how-to-show-divider-between-spinner-items

PR #2107, opened 2018-05-15
Feature v1.1 login activity icon

### LINKED ISSUE #2100: Change splash screen logo by connect one
Just like we did in HNQIS

idelcano:
Closes #2100


app/src/ereferrals/java/org/eyeseetea/malariacare/strategies/LoginActivityStrategy.java
```diff
@@ -74,6 +78,14 @@ public LoginActivityStrategy(LoginActivity loginActivity) {
         mPullUseCase = new PullUseCase(pullController, asyncExecutor, mainExecutor);
     }
 
+
+    private void replaceDhisLogoToHNQISLogo(LoginActivity loginActivity) {
```
xurxodev: @idelcano I think that replace logos in runtime is not the better option, please you intent use only layouts, variants and resources for this issue and we should not replace in runtime resources because is not the best optime option

xurxodev (changes requested):
@idelcano i have found some code to review, please look my comments in the code. Thanks

xurxodev (approved):
thanks @idelcano

Issue #2112, opened 2018-05-16
Refactor retrieval of the appropriate program to open the survey


ifoche:
The app opens the survey in 3 different ways, they all need to take into account that the program to be opened is the one in preferences (accessed through the programRepository)

PR #2115, opened 2018-05-16
Feature fix bug survey program

### LINKED ISSUE #2111: Make survey open the proper program in every open call in the app
Pay special attention to refactor the code added in PR #2110

idelcano:
Closes #2111


ifoche:
I'm having problems compiling the code...I think it has to do with my last AStudio upgrade. Please @xurxodev if you can take care of this PR, that would be perfect. From the point of view of the code, the code repetition is ok, there's already the PR #2112 to deal with that. I was able to compile it before merging with v0.4_cnm, and I found a bug (VMW surveys partially filled in). To discard any problem in the final code, I updated the code, but then I got stuck with this compilation problem. So please, play a little bit changing to another app and coming back again to CNM while in the middle of a survey, just to be sure that we're not missing anything here. Thank you!

xurxodev (approved):
thanks @idelcano

PR #2116, opened 2018-05-17
Refreshing only webviews only first time and when user clicks button …

### LINKED ISSUE #2102: Refreshing webviews
This task implies:
- [ ] Remove automatic refresh of webviews (they're refresing periodically every minute or something similar)
- [ ] Cache the webviews, so when they're accessed the first time, they're not reloaded on tab change
- [ ] Add floating button (just like the + button on the first tab) to manually refresh. This button should be greyed out when the app is offline

manuelplazaspalacio:
Close #2102


xurxodev (changes requested):
@manuelplazaspalacio gradle-wrapper.properties file in this issue should not change

xurxodev (approved):
thanks @manuelplazaspalacio

PR #2117, opened 2018-05-17
Feature custom dropdown with styles

idelcano:
Related #2105
trying to use styles but dividers doesn't work with spinner in dialog mode.


xurxodev (changes requested):
thanks @idelcano. I refactored the code to set dividers with style without use strategy. I will create a new issue for does not use code, only layout and styles to set the background to selected also.

PR #2122, opened 2018-05-22
fix strange effect closing surveyfragment

### LINKED ISSUE #2119: When going back from a voucher to the dashboard the myanmar arrows show in the change effect
If you are creating a voucher and you press back, in the visual effect that takes you from the survey fragment to the unsent list, you can see the myanmar arrows to go forward and backwards.

idelcano:
Closes #2119

### DIFF
--- a/app/src/cambodia/java/org/eyeseetea/malariacare/strategies/DropdownAdapterStrategy.java
+++ b/app/src/cambodia/java/org/eyeseetea/malariacare/strategies/DropdownAdapterStrategy.java
@@ -1,0 +1,13 @@
+package org.eyeseetea.malariacare.strategies;
+
+import android.view.View;
+import android.widget.ListView;
+import android.content.Context;
+
+import org.eyeseetea.malariacare.R;
+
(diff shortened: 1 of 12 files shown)


xurxodev (approved):
thanks @idelcano, I have modified your solution because add code is not necessary, only modify layouts.
If we include back_button layout invisible, add code to remove removeAllViews is not necessary and Its most optimal solution.
We should avoid solutions adding code if exists solution modify layout, styles, resources etc..

idelcano:
@xurxodev I tried your solution before but I searched other solution
because the back buttons is not visible but the empty layout yes(you can
see the header and the submit button with empty content).
If you go back you see the survey layout empty before the assess tab. This
is the reason to add the code and I think is better. The empty survey.xml
should not be visible when the user press back.

xurxodev:
@idelcano ok but we must always avoid code solutions for interface problems if there is one through layout, resources or styles because they are much cleaner solutions and do not dirty the code. In any case, what you are saying would be another issue, not what the client has reported.

PR #2124, opened 2018-05-23
Adding accuracy to de surveys pull.

### LINKED ISSUE #2118: GPS accuracy
- [ ] Retrieve the GPS accuracy: https://developer.android.com/reference/android/location/Location.html#getAccuracy()
- [ ] Send it to the WS at the same level as "coordinate" and with the key "accuracy"

manuelplazaspalacio:
Close #2118


xurxodev (approved):
thanks @manuelplazaspalacio

PR #2143, opened 2018-06-04
Feature refactor in progress surveys deletion system

### LINKED ISSUE #2120: Refactor in progress surveys deletion system
Currently we're deleting every single survey in progress each time we access dashboard. This was a measure that we introduce to avoid blanks long time ago, but we demonstrated also long time ago that blanks came from other blank sources, different than this. The goal of this issue is to think a bit if this code is strictly needed, or if it's simply something that make things more complicated without adding any significant value. For example...In which situations we will need this? Can we simply move it to EyeSeeTeaApplication on create and just manage the event when the app crashes while the user was fillling in a form?

idelcano:
Closes #2120


ifoche (approved):
In order to avoid disturbing tomorrow's tests and because it's not a strong reason to stop this needed refactor, I'm not asking for any change and accepting it, but please @idelcano, could you have a special look tomorrow to the new survey graphical performance? I have the impression that executing this use case in the main thread before creating the new survey is affecting the performance. In fact, I now cannot see the screen effect when the new survey is created. Thanks!

PR #2144, opened 2018-06-04
remove eula from cnm

### LINKED ISSUE #2141: EULA shows PSI
Please remove that, the EULA is the same as PSI app, as if the app belongs to PSI. Please remove

idelcano:
Closes #2141


ifoche (approved):
thanks @idelcano

PR #2148, opened 2018-06-07
Maintenance store upgrade metada date

### LINKED ISSUE #2139: Store time/date when upgrade took place and show it in about us
Place it next to config version, so we can easily see when did that update took effect

manuelplazaspalacio:
Close #2139

### DIFF
--- a/app/src/ereferrals/java/org/eyeseetea/malariacare/data/sync/exporter/ConvertToWSVisitor.java
+++ b/app/src/ereferrals/java/org/eyeseetea/malariacare/data/sync/exporter/ConvertToWSVisitor.java
@@ -44,10 +44,12 @@ public class ConvertToWSVisitor implements IConvertToSDKVisitor {
 
     private SurveyContainerWSObject mSurveyContainerWSObject;
     private String language;
+    private Context mContext;
 
-    public ConvertToWSVisitor() {
+    public ConvertToWSVisitor(Context context) {
         IDeviceRepository deviceDataSource = new DeviceDataSource();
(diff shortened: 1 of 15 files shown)


app/src/ereferrals/java/org/eyeseetea/malariacare/data/sync/exporter/ConvertToWSVisitor.java
```diff
@@ -44,10 +44,12 @@
 
     private SurveyContainerWSObject mSurveyContainerWSObject;
     private String language;
+    private Context mContext;
 
-    public ConvertToWSVisitor() {
+    public ConvertToWSVisitor(Context context) {
```
ifoche, line 49: @manuelplazaspalacio instead of making the visitor depend on the context, could you please just add the metadata update date between brackets just after the config file version number? Something like "Config Version: 4 (06/10/2018)". This way you don't need to pass the context through all the classes just to translate one string.

ifoche (changes requested):
@manuelplazaspalacio I've requested one change, but also there's a functional change I would like to ask for. If you just install on top of the previous version, the config update date that it shows is 1 jan 1970. Could we simply show a "-" when there's no previous registry of that date or setup the upgrade time as that date? Thanks!

ifoche (approved):
thanks @manuelplazaspalacio. Just for the record, we finally decided to keep context there because it was already in use in the class, but using the static context. So we refactored it to make it use the context passed in the constructor.

PR #2151, opened 2018-06-09
increase x5 ereferals timeout

### LINKED ISSUE #2136: Push to send record triggered again (when record is received by WS)
Alpha tester report:

This looks like a cause of duplicates. I attach a phone database to look at this in details.

idelcano:
Closes #2136 
The problem is server timeout.
I increased the app timeout to solve it waiting for the server response.
But i think we should create a quarentine system to ereferals to avoid duplicates. But we would need a call to ask for the surveys info in the webservice.

to replicate this bug :
User: MZ_TEST_IPC
server: https://clone.psi-mis.org
webservice: https://apps.psi-mis.org/eRefWSTrain/
web url(in settings): https://apps.psi-mis.org/connect-train/

### DIFF
--- a/app/src/ereferrals/java/org/eyeseetea/malariacare/data/sync/exporter/eReferralsAPIClient.java
+++ b/app/src/ereferrals/java/org/eyeseetea/malariacare/data/sync/exporter/eReferralsAPIClient.java
@@ -36,7 +36,7 @@ public class eReferralsAPIClient {
     private SurveyApiClientRetrofit mSurveyApiClientRetrofit;
     private OkHttpClient mOkHttpClient;
     public String mBaseAddress;
-    private final int DEFAULT_TIMEOUT = 10000;
+    private final int DEFAULT_TIMEOUT = 50000;
 
     public eReferralsAPIClient(String baseAddress) throws IllegalArgumentException {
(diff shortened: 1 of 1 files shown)


ifoche (commented):
@idelcano can you explain why increasing the timeout by 5 avoid those duplicates to happen? Let's talk about this tomorrow

idelcano:
@ifoche the app waits for the server response but the server is slow and
the app throw a timeout exception, but the server save the voucher anyway..
The app tries again, and again... and only stops when the server response
is more faster than the app timeout. I avoid this problem only in this case
increasing the timeout.

The good solution is create a quarantine system. But this option needs a
webserver update to recovery the pushed vouchers to identify the duplicates.

ifoche:
Ok, thank you for your answer. 2 things:
1) WS is being updated to allow verification of UIDs, so it will shortly be possible to establish a quarantine system as we have in the other apps
2) If the problem is that we save the voucher even when the server answer with a timeout...can't we just wait for the server response before saving the voucher state change?

ifoche (approved):
According to our conversation this morning, I'm approving this PR and have communicated with alpha testers that there's a response time problem in the server they need to fix

PR #2154, opened 2018-06-12
Feature integrate elements library

### LINKED ISSUE #2147: Connect connect with elements
The platform is: https://www.discoverelement.com/

They have an installable SDK here:
https://github.com/Element1/element-android-examples/tree/master/orbit-lite-integration-example 
More examples integrating with it here: https://github.com/Element1/element-android-examples/tree/master/orbit-lite-integration-example
More documentation: https://element1.github.io/element-android-examples/

The expected behavior is to create an intent against that sdk whenever a paper voucher is generated passing the autentication parameters to element.

idelcano:
Closes #2147


app/src/ereferrals/java/org/eyeseetea/malariacare/data/remote/ElementController.java
```diff
@@ -0,0 +1,50 @@
...
+
+import android.app.Activity;
+import android.content.Intent;
+import android.util.Log;
+
+import com.element.utils.ElementSDKManager;
+
+import org.eyeseetea.malariacare.domain.boundary.IElementController;
+
+public class ElementController implements IElementController, ElementSDKManager.ElementActivityListener, ElementSDKManager.EnrollListener {
```
ifoche: I think the ElementController should implement an interface of ExternalVoucherRegistry or something similar. The idea would be that Element is a technology that is receiving the post-send action of a voucher, and this controller just pass the ID to element, but if later another technology want to receive that voucher ID and do a different action, it would only need to implement the same interface.

app/src/ereferrals/java/org/eyeseetea/malariacare/data/remote/ElementController.java
```diff
@@ -0,0 +1,50 @@
...
+    private Activity mActivity;
+    private Callback mCallback;
+
+    public ElementController(Callback callback, Activity activity){
+        mCallback = callback;
+        mActivity = activity;
+    }
+
+    @Override
+    public void sendVoucher(String voucherUId) {
```
ifoche: This is a bit confusing. Actually, it's not sending a voucher, but the UID to Element, for it to deal with it. I think the interface methods naming should be changing according to my previous comment, and do what it's intended to do. Maybe the interface could be prepared to receive the whole voucher (domain entity) and element controller just take what it needs (voucher id) and pass it to element? Something like a post-send voucher action, that in this case, for element, is just send the voucher id.

app/src/ereferrals/java/org/eyeseetea/malariacare/domain/boundary/IElementController.java
```diff
@@ -0,0 +1,9 @@
+package org.eyeseetea.malariacare.domain.boundary;
```
ifoche: As explained before, I think we need to generalize much more this interface to any post-send action, and not only for Element.

app/src/ereferrals/java/org/eyeseetea/malariacare/domain/usecase/ElementOnActivityResultUseCase.java
```diff
@@ -0,0 +1,31 @@
+package org.eyeseetea.malariacare.domain.usecase;
```
ifoche: It sounds a bit strange to me this Use Case and the following one. Actually, they just execute in a separate thread one thing. Shouldn't this be part of the Controller? This is like managing the result of the post-send action, so I would expect the controller to have some error treatment embedded. Can you think again how t refactor these use cases?

app/src/ereferrals/java/org/eyeseetea/malariacare/strategies/DashboardActivityStrategy.java
```diff
@@ -526,12 +557,69 @@ public void reloadAVFragment() {
...
+                    mDashboardActivity.showException(mDashboardActivity, "", String.format(
+                            mDashboardActivity.getResources().getString(R.string.give_voucher),
+                            voucherUId), onClickListener);
+                }
+            });
         }
     }
 
+    @NonNull
+    private DialogInterface.OnClickListener createOnClickListenerToSendVoucherToElement(final String voucherUId) {
```
ifoche: There is a lot of Element-dependent logic in the dashboard activity. It's probably the same "naming-convention" problem, but I would expect the dashboard to talk about post-actions that are needed, not about a concrete technology.

app/src/main/java/org/eyeseetea/malariacare/EyeSeeTeaApplication.java
```diff
@@ -59,6 +60,9 @@
     public void onCreate() {
         super.onCreate();
         Log.d(TAG, "onCreate");
+        if(BuildConfig.elementSDK) {
```
ifoche, line 65: This init shouldn't happen directly here (see the dependency that's created via the import with the ElementSDKManager). Instead of that, I would rely on the ElementController for that purpose. The better would be something like a plugin initialization where all the Controllers that need to be initialized simply subscribe and are initialized here, but if that's too much, at least it's better to have a dependency with our generic controller and not with the element SDK.

ifoche (changes requested):
@idelcano thank you. From a functional perspective, it's working, but I would like some pieces of code to be refactored when possible. Thanks!

idelcano:
@ifoche  it`s done but we could not avoid to use two steps (a first step to send the voucher uid, and second step, who starts from the activity to treat the activity result)

ifoche (approved):
thanks @idelcano

PR #2179, opened 2018-06-14
Solved bug not showing the sending surveys in the list of DashboardAc…

### LINKED ISSUE #2175: SENDING surveys don't appear in the unsent list
Please look at the following PR: #2163

manuelplazaspalacio:
Close #2175


ifoche (changes requested):
@manuelplazaspalacio if I create a quick survey not tested -> patient refused it is showing me 2 entries in the dashboard now...probably it's related to the stock survey. I think you're missing the program filter that we need in CNM and not in Connect. Can you have a look to it? Thanks!

FYI: 
screenshot_20180614-201832 1

ifoche (approved):
thanks @manuelplazaspalacio

PR #2186, opened 2018-07-04
Upgrade OkHttp to version 3.8 and resolve conflicts

### LINKED ISSUE #2184: Change server URL
The URL is this one:
https://sandbox.cnm-mis.org/

We've already tested it in the past, and the App was producing by that time an exception that seemed to be related to an incompatiblity between that precise nginx server and okHttp library. We need to have a look in deep to this, in order to investigate if we can find a quick workaround to solve the problem app side, or if it's something that can only be solved by an nginx server upgrade (or a okHttp version upgrade...which I think is too much work)

xurxodev:
### :pushpin: References
* **Issue:** close #2184 
* **Related pull-requests:** 

Change default url server from https://dhis2.asia/cnm_android to https://sandbox.cnm-mis.org/ and solve exceptions with new url.

The problem with new url server is incompatibility of nginx version with old OkHttp versions
https://trac.nginx.org/nginx/ticket/1397

I have upgraded OkHttp version to 3.8 and resolve conflicts because from Okhttp 2.x versions to 3.x version the API has changed.

- [x] **Use case 1:** Test with HC org unit 
    - [x] realize login and assign HC org unit
    - [x] create survey
    - [x] wait to push and review uploaded surveys in new server

- [x] **Use case 2:** Test with VL org unit 
    - [x] realize login and assign VL org unit
    - [x] create survey
    - [x] wait to push and review uploaded surveys in new server

- [x] **Use case 3:** Test autoconfigure 
    - [x] Include IMEI in the phone number field of an OU
    - [x] Launch the app and wait for it to autoconfigure that OU
    - [x] create survey
    - [x] wait to push and review uploaded surveys in new server

### :art: UI changes?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -299,7 +299,7 @@ dependencies {
     compile "com.opencsv:opencsv:${libs.opencsvVersion}"
     compile "com.google.android.gms:play-services:${libs.gmsPlayVersion}"
     compile "com.google.guava:guava:${libs.guavaVersion}"
-    compile "com.squareup.okhttp:okhttp:$rootProject.okhttpVersion"
+    compile "com.squareup.okhttp3:okhttp:${libs.okhttpVersion}"
     compile "com.squareup.okhttp3:logging-interceptor:${libs.okhttpVersion}"
     compile "com.googlecode.libphonenumber:libphonenumber:${libs.libphonenumberVersion}"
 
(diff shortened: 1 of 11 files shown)


ifoche (approved):
thanks @xurxodev

xurxodev:
@manuelplazaspalacio The use case to test this issue has no sense in malaria because does not exists new server, please avoid copy and paste from other PR descriptions so that using a new template is worthwhile

PR #2196, opened 2018-07-13
Force loging with hardcoded user before push events from sdk

### LINKED ISSUE #2194: Cases do not upload after network interruption in VMW app
That only happens when the login in the VMW app is done with Username and Password, the next steps will help recreate the issue:

1. Login to the desired VMW village with the username and password
2. Exit app and then turn off all data on your device
3. Return to the app with data still off (this is a common use case for VMWs) and report a case. As expected it will not upload since there is not data
4. Exit app and turn on the phone’s data
5. Return to the app to upload the case. You will notice it will not upload. In some cases the case will flicker where it would disappear and then return. I waited 10+ minutes and the case would not upload. You can try adding another case at this point but it did not upload for me either

idelcano:
Closes #2194


ifoche (approved):
thanks @idelcano

PR #2205, opened 2018-07-13
Maintenance new autocomplete control type

### LINKED ISSUE #2189: New autocomplete control type
We need to generate a new autocomplete control type. It would be essentially the same as the dropdown one, but with a textfield on top of it. While the user writes, the options rendered would filter in real-time showing only those containing that exact string at any position.

manuelplazaspalacio:
Close #2189


xurxodev (changes requested):
@idelcano This solution applied by @manuelplazaspalacio is not correct to the user experience level according to the issue description. Can you review it?

app/src/main/java/org/eyeseetea/malariacare/views/question/multiquestion/DropdownWithFilterMultiQuestionView.java
```diff
@@ -0,0 +1,290 @@
...
+       //}
+ //   }
+
+    @Override
+    public boolean hasError() {
+        return false;
+    }
+
+    private void init2(final Context context) {
+      //  inflate(context, R.layout.multi_question_tab_dropdown_row, this);
```
xurxodev: Remove this unused method

app/src/main/java/org/eyeseetea/malariacare/views/question/multiquestion/DropdownWithFilterMultiQuestionView.java
```diff
@@ -0,0 +1,290 @@
...
+    public void setEnabled(boolean enabled) {
+        textViewButton.setEnabled(enabled);
+    }
+
+    @Override
+    public void setHelpText(String helpText) {
+
+    }
+
+   //@Override
```
xurxodev: Remove commented code

app/src/main/java/org/eyeseetea/malariacare/views/question/multiquestion/DropdownWithFilterMultiQuestionView.java
```diff
@@ -0,0 +1,290 @@
...
+                            }
+                        }
+                    }
+                });
+                editText.requestFocus();
+            }
+        });
+    }
+
+    private void filter(ListView listView, String s) {
```
xurxodev: rename parameter s to more readable name

app/src/main/java/org/eyeseetea/malariacare/views/question/multiquestion/DropdownWithFilterMultiQuestionView.java
```diff
@@ -0,0 +1,290 @@
...
+        selectedOptionDB = valueDB.getOptionDB();
+        if (BuildConfig.validationInline) {
+            if(textViewButton.getText()==null || textViewButton.getText().toString().isEmpty()) {
+                Validation.getInstance().addinvalidInput(header, getContext().getString(
+                        R.string.error_empty_question));
+            }
+        }
+    }
+
+    private void init(final Context context) {
```
xurxodev: this method is big, please refactor to more readable code

xurxodev (changes requested):
@idelcano I have found something to review.
- We are not using spinner but for the user should seem it, I think can to help if we add one arrow icon to right like  in spinner
- When I type text the filter work but if I remove the filter all options should appear again
- There are some improvements to realize in new DropdownWithFilterMultiQuestionView to convert it more readable code, I have inserted comments in code

xurxodev (changes requested):
@idelcano now the arrow is perfect, code of DropdownWithFilterMultiQuestionView has no commented code but I have found an error when I select a program from new control, the selected program is assigned to another field
[Video](https://drive.google.com/file/d/1791q-1A8dXrbAJ-Z4BGKCVo9Fq1xUjGv/view?usp=sharing)

xurxodev (commented):
This error only occurs if I have another field like autocomplete. In my scenario, I have the program and the PhoneOwnership field with control type equal to AUTOCOMPLETE

xurxodev (approved):
thanks @idelcano

Issue #2188, opened 2018-07-10
Add regexp validations


ifoche:
Any question should accept regexp validations. In order to do this, at least we would need:
- [x] Parse the new keys in the configuration file that would contain this information:
```
{ "dataPointRef": "age",
  "poTerm": "ipc_issueEntry_q_age",
  "defaultName": "Idade",
  "controlType": "INT",
  "validationRegex": "^(\d{2})$",
  "validationPoTerm": "some_error_msg_ref"
}
```
The "validationRegex" key would contain the expresion to apply, while the "validationPoTerm" would contain the POEditor key for the error in case validation doesn't pass

- [x] Store this information in the DB. Probably we would need to slightly modify the DB to allow the storage of this information. Ideally, the validation would be parsed in a separate table linked to the Question one.

- [x] Modify the validation logic so that the validation is applied in case there's a regexp associated to the question

PR #2207, opened 2018-07-16
Feature add regexp

idelcano:
Closes https://github.com/EyeSeeTea/pictureapp/issues/2188

@xurxodev 
To test this issue:

For the user 8001 and clone server:

copy the content on https://clone.psi-mis.org/api/dataStore/Connect_config/dc@TZ@v4 

Modify https://clone.psi-mis.org/api/dataStore/Connect_config/dc@TZ@v4 adding 
"  ,
"validationRegex": "^(\\d{2})$",
  "validationPoTerm": "some_error_msg_ref"
"
in some question (for example, firstName, and adolescentDaughters).
send this modified json with postman using "put" method.
PD: you can test the translations of the PO terms using a term already translated, for example, 
  "validationPoTerm": "ipc_issueEntry_q_adolescentDaughters"


xurxodev (changes requested):
@idelcano after talk with @ifoche, Regex info for question is a relation one to one, then is better that we don't create separate table to store this info because this one is not necessary.
Please can you remove this new table and to add regex fields in question?
Respect to domain this info should be in question also

xurxodev (approved):
thanks @idelcano

PR #2208, opened 2018-07-18
Feature check metadata on soft login setting

### LINKED ISSUE #2192: Create check metadata on soft logins settings
We need to create a new setting reading:
Check metadata updates on soft logins

This setting would only be visible if developers options is activated. Once checked, the app wouldn't check any metadata or POEditor terms update during a soft-login. The idea is to boost the login process when no updates are done.

idelcano:
Closes #2192


xurxodev (changes requested):
@idelcano new settings 'Check metadata updates on soft logins' should be verified only in soft login but It's verified always also with normal login. Can you review it?

xurxodev (approved):
thanks @idelcano

PR #2209, opened 2018-07-18
Feature authorize extra keys

### LINKED ISSUE #2190: Authorize extra keys on WS responses
Each time the WS answers with a key that we don't know how to parse, the parsing crashes. We need to add anonymous builders or something like that to allow the WS response to contain other keys, always while keeping at least the minimum ones that we're currently expecting.

idelcano:
Closes #2190 
I could not replicate this problem, but I think this could solve the problem. I also fixed the WS response test.


xurxodev (approved):
thanks @idelcano 
I have reproduced the extra key error, creating a test against eReferralsAPIClient. 
I have refactored some tests to use modern testing API to retrieve context

PR #2231, opened 2018-08-13
Remove delay before go to next question

### LINKED ISSUE #2230: Refactor oriented to reduce transition time between screens
We need to spend some time refactoring the code to speed up the screens transitions. To do it properly, probably we would need to touch the workflow engine, but there's not too much time to do it, so probably we need to invest a little time analysing where we could refactor to improve the performance. Even if we end up with a very little almost null performance improvement, given that it's not too much time, we can accept it.

xurxodev:
### :pushpin: References
* **Issue:** close #2230 
* **Related pull-requests:** 

Reduce transition time between survey questions

I have deleted a delay before to navigate to the next question.

- [ ] **Use case 1:** Realize login against an HC program and create a new survey, the navigation should be faster
- [ ] **Use case 2:** Realize login against a VL program and create a new survey, the navigation should be faster

### :art: UI changes?


ifoche (approved):
thanks @xurxodev

PR #2242, opened 2018-08-23
Remove question on reported within 24hrs

### LINKED ISSUE #2241: Remove question on reported within 24hrs
Note that four different screens currently point to this screen, so probably they are separate questions (Treatment, Reason for referral, Microscopy: negative and RDT: negative), and they will all need to be redirected to the 'Date of case' screen.

xurxodev:
### :pushpin: References
* **Issue:** close #2241  
* **Related pull-requests:** 

Remove question on reported within 24hrs

Edit csv files

- [ ] **Use case 1:** Realize login against an HC program and create a new survey and verify after this questions-options: either Treatment, Reason for referral, Microscope- negative and RDT- negative the next question is Date of case
- [ ] **Use case 2:** Realize login against a VL program and create a new survey and verify after this questions-options: either Treatment, Reason for referral, Tested- negative and RDT- Stock out the next question is Date of case

### :art: UI changes?


ifoche (approved):
thanks @xurxodev

PR #2244, opened 2018-08-24
Feature add patient number

### LINKED ISSUE #2237: Add (after Patient Name) simple integer box for Patient Number
Data element has already been created (KH_MCS_PTO - Patient number)

xurxodev:
### :pushpin: References
* **Issue:** close #2237   
* **Related pull-requests:** 

Add (after Patient Name) simple integer box for Patient Number

Edit csv files

- [ ] **Use case 1:** Realize login against an HC program and create a new survey and verify after Patient Name question the next question is Patient Number. Verify push is ok.

- [ ] **Use case 2:** Realize login against a VL program and create a new survey and verify after Patient Name question the next question is Patient Number. Verify push is ok.

### :art: UI changes?


ifoche (approved):
thanks @xurxodev. I'm accepting the PR as they didn't specify if the integer would have a limit less than 99 or there's no limit. You're using POSITIVE_OR_ZERO_INT, and it's perfect, but it seems that at some point we limited it to represent the age question. Probably in the future it would be great to create a separate type for AGE (as we have for AGE_MONTH_NUMBER) and remove the 2-digits limit that we have for POSITIVE_OR_ZERO_INT.

PR #2276, opened 2018-09-27
Feature merge v0.4 cnm into development

### LINKED ISSUE #2275: Bring last CNM 0.4 changes into development

idelcano:
### :pushpin: References
* **Issue:** close #2275

Bring the last cnm 0.4 changes into development

I only merge the v0.4_cnm branch into development

- [x] **Use case 1:** Check "For HC app, change in first screen "Not tested" to "Referred without tested" and keep same workflow"

- [x] **Use case 2:** Check https://github.com/EyeSeeTea/pictureapp/pull/2231

### :art: UI changes?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -220,7 +220,7 @@ android {
         }
         cnm {
             applicationId 'org.eyeseetea.surveillance_cnm_dev'
-            versionName "0.4.4"
+            versionName "0.4.5"
             buildConfigField "boolean", "activeMonitoringFragment", "false"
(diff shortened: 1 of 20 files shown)


xurxodev (changes requested):
@idelcano  I have found a bug testing this branch, I don´t know if the problem exists on release branch too but I think that we can fix it:
- When creates a new survey always appear two new surveys in the unsent list, the created one and other in progress

xurxodev (approved):
thanks @idelcano

PR #2279, opened 2018-10-05
Feature update elements

### LINKED ISSUE #2253: Elements update
A new version is available: https://rink.hockeyapp.net/apps/458abb63bfb442b0afc8989fd0e8b853

After the update, opening the element sdk does not work anymore:
"Unauthorized used of the SDK".

See screencast: https://drive.google.com/open?id=1yKqjgOkdxujhCtabfqltpRNQOvIkuGGz

idelcano:
### :pushpin: References
* **Issue:** close #2253 
* **Related pull-requests:** 

Update sdk to avoid the "Unauthorized used of the SDK" error.

I updated the library and update the new required key.
I added to a line into ereferrals androidManifest to allow up the new required minimal target by orbit (api 21), and exclude other variants to compile orbit.

Note: I found a possible bug or configuration problem in the element servers.
The element`s user is created in the app and orbit library but i didn't find the user in the element dashboard, and i found the follow bug in the logcat:
```
com.amazonaws.services.s3.model.AmazonS3Exception: 
The bucket you are attempting to access must be addressed using the specified endpoint.
 Please send all future requests to this endpoint. 
(Service: Amazon S3; Status Code: 301; Error Code: PermanentRedirect; Request ID: E17935EE78A725F2), 
S3 Extended Request ID: 7k49jbx+Zzj1KeW8WY5p46VNww1orY7pxikIGK8tXMb4UP7m7aTSjgfdx5jlVCKqfcWy9VFxns8=
```
I think this bug is relate with the bucket regions:
https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region

Note: Element app should be installed:

https://rink.hockeyapp.net/apps/458abb63bfb442b0afc8989fd0e8b853

- [ ] **Use case 1:** Go to the configuration, activate the developer options and activate element. You create a survey, the element application will be launched and you can to create a new user.

### :art: UI changes?


xurxodev (approved):
thanks @idelcano

PR #2273, opened 2018-09-21
Feature remove dhis2 server

idelcano:
### :pushpin: References
* **Issue:** close #2250  
* **Related pull-requests: https://github.com/EyeSeeTea/pictureapp/pull/2270 ** 

### :tophat: What is the goal?

 Remove DHIS2 server URL setting on ereferrals.

 Change url in login to https://apps.psi-mis.org  

 Change login server settings from DHIS2 preference to use Webservice preference

### :memo: How is it being implemented?

Note: i didn't change the default url used in datastore calls "https://data.psi-mis.org", https://apps.psi-mis.org doesn't work and we are waiting for other endpoint. But the server preference in login use the webservice preference url now (https://apps.psi-mis.org/eRefWSProd/).

I removed dhis server preference from SettingsActivity in ereferrals and move from main to the other variants.

I create a string resource named "server_url_key" in each variant that contains the server url preference key, in the case of erreferals this key contains the webservice key in the other variants the old dhis_url key. 

I replace all the dhis_url references for the common code by server_url_key.

I changed the dhis default value, and i renamed the resource DHIS_DEFAULT_SERVER by DEFAULT_SERVER_URL to use this variable  in all the variants. In ereferrals that url is the webservice("@string/ws_base_url")

I merged the https://github.com/EyeSeeTea/pictureapp/pull/2265 PR to avoid login crash.

### :boom: How can it be tested?

Remember change server url to https://apps.psi-mis.org/eRefWSDev3/ 

- [ ] **Use case 1:** Check the login works and login server url is on webService url on settings.

### :floppy_disk: Requires DB migration?

- [x] Nope, we can just merge this branch.
- [ ] Yes, but we need to apply it before merging this branch.
- [ ] Yes, it's already applied.

### :art: UI changes?

- [x] Nope, the UI remains as beautiful as it was before!
- [ ] Yeap, here you have some screenshots-

Issue #2250, opened 2018-09-04
Remove DHIS2 server URL


xurxodev:
- [ ] Remove DHIS2 server URL setting
- [ ] Change url in login to https://apps.psi-mis.org
- [ ] Change login server settings from DHIS2 to Webservice

idelcano:
@xurxodev

this task : 
"Remove DHIS2 server URL setting" is done in #2249 (but i only hide the DHIS2 server URL preference instead of remove it from the code, it is used in the main code). 

And the second task: "Change url in login to https://apps.psi-mis.org or remove EditText?"
that I have to do? Only change the default url to https://apps.psi-mis.org ?.
The default url is used by #2249 into new Program Configuration URL as default url.

PR #2281, opened 2018-10-05
Fix server settings bug

### LINKED ISSUE #2278: Fix Login process from settings
Require #2250 because is where the bug is introduced:
 - [ ] ServerURL is showed only when developer options preference is true. This behaviour is wrong, in CNM URL server should be visible always.
- [ ] When I realize login changing url server, then I can not select org unit because always request login again. I can not complete login process (login + select org unit)

idelcano:
### :pushpin: References
* **Issue:** close #2278 
* **Related pull-requests: https://github.com/EyeSeeTea/pictureapp/pull/2273 ** 

 Fix the problems::
-[x] ServerURL is showed only when developer options preference is true. This behaviour is wrong, in CNM URL server should be visible always.
-[x] When I realize login changing url server, then I can not select org unit because always request login again. I can not complete login process (login + select org unit)

I remove strange code who remove the server setting.

And i removed a old variable.

- [ ] **Use case 1:** Login with cnm and with ereferals, go to settings and change url.

### :art: UI changes?


xurxodev (commented):
@idelcano this cnm branch has a connect branch as base, It's correct?

idelcano:
@xurxodev  no, this Branch starts from https://github.com/EyeSeeTea/pictureapp/issues/2250 as is required by the issue https://github.com/EyeSeeTea/pictureapp/issues/2278

This bug is introduced in 2250 and doesn't exist in other Branch.

xurxodev (changes requested):
@idelcano please wait until that issue #2250 is merged because is required by issue and then review this issue changing base to development

xurxodev (approved):
thanks @idelcano

PR #2295, opened 2018-10-12
added local date time of the language request to save if a donwloaded…

### LINKED ISSUE #2285: nullpointerexception on language translation use case
## User Report
when device is not configured in English. variable lastupdated is null

## ScreenShot
file
## Device Info
```
Time Stamp: 2018-10-12T08:00:07 UTC
App Version: 1.2.1 (59)
Install Source: Unknown
Android Version: 8.1.0 (27)
Device Manufacturer: BQ
Device Model: AQUARIS X PRO
Display Resolution: 1920x1080
Display Density (Actual): 408dpi
Display Density (Bucket) Unknown
---------------------

```

idelcano:
… language doesnt have date

### :pushpin: References
* **Issue:** close #2285 

Fix language crash.

Some languages doesn't have update Date. 
I think that languages(spanish and french) was not filled in POeditor.

I created a Date variable with the local date before the language download, and i setted this date if the language doesn't have date.
In that case, if the language will be update in the server, the new date will be newer than the local date and the app will update the terms. 

- [ ] **Use case 1:** Login with spanish or french as language system, the logcat doesn't show crash. 
Note: The spanish/french terms are empty.

### :art: UI changes?


xurxodev (approved):
thanks @idelcano

PR #2317, opened 2018-10-26
Feature add slow load of village question

### LINKED ISSUE #2309: HC Program: Very slow transition from Test Result RDT question to Village of Residence question

idelcano:
### :pushpin: References
* **Issue:** close #2309
* **Related pull-requests:** 

HC Program: Very slow transition from Test Result RDT question to Village of Residence question
Make it faster again.

We added some validations in ereferrals. The new validations requires the mapping of QuestionDB to Question.
In the case of Org unit tree quesiton the validations aren't required, and the map from questionDb to Question is so slow because the question has about 10.000 options and option attributes.

I added a check to know if the question requires that validation to avoid that unnecesary mapping in the case of the org unit tree question.

- [ ] **Use case 1:** The load of village question should be shown in less than 5 seconds.

### :art: UI changes?


xurxodev (approved):
@idelcano I have approved this PR but How we have talked, this code of the app is a little strange and dangerous. I have created two issues to refactor on the future:
https://github.com/EyeSeeTea/pictureapp/issues/2318
https://github.com/EyeSeeTea/pictureapp/issues/2319

PR #2316, opened 2018-10-26
Solve bug that makes app crash when answering som questions.

manuelplazaspalacio:
### :pushpin: References
* **Issue:** close  #2308 close #2310 
* **Related pull-requests:** 

### :tophat: What is the goal?

Solve bug that makes app crash when answering survey questions.

### :memo: How is it being implemented?

In single questions  initializing the answer edit text from AKeyboardQuestionView and in AKeyboardQuestionView checking that this edit text isn't null before using it.

### :boom: How can it be tested?

- [ ] **Use case 1:** Login with HC program and test all posible paths in surveys.
- [ ] **Use case 2:** Login with Vl program and test all posible paths in surveys.

### :floppy_disk: Requires DB migration?

- [x] Nope, we can just merge this branch.
- [ ] Yes, but we need to apply it before merging this branch.
- [ ] Yes, it's already applied.

### :art: UI changes?

- [x] Nope, the UI remains as beautiful as it was before!
- [ ] Yeap, here you have some screenshots-

Issue #2314, opened 2018-10-26
Crash after crash


EyeSeeTeaBotTest:
## User Report
after a crash start a new survey and...

## ScreenShot
![file](https://raw.githubusercontent.com/EyeSeeTeaBotTest/snapshots/master/android_screenshot9201877916709665784.jpg)
## Device Info
```
Time Stamp: 2018-10-26T07:28:05 UTC
App Version: 0.5.0 (61)
Install Source: Unknown
Android Version: 8.1.0 (27)
Device Manufacturer: BQ
Device Model: AQUARIS X PRO
Display Resolution: 1920x1080
Display Density (Actual): 408dpi
Display Density (Bucket) Unknown
---------------------

```

ifoche:
BTW I didn't copy the crash details, but just in case:
```
`ProgramProgramRelation` AS `pp` ON `pg`.`id_program`=`pp`.`id_aux_program` WHERE `pp`.`id_main_program`=3
2018-10-26 09:27:24.131 25949-25949/org.eyeseetea.surveillance_cnm_dev D/.SurveyFragment: onCreateView
2018-10-26 09:27:24.167 25949-25949/org.eyeseetea.surveillance_cnm_dev I/System.out: Exception info: Error loading new survey, null navigation controllerAttempt to invoke virtual method 'void org.eyeseetea.malariacare.layout.adapters.survey.navigation.NavigationController.first()' on a null object reference
2018-10-26 09:27:24.167 25949-25949/org.eyeseetea.surveillance_cnm_dev W/System.err: java.lang.NullPointerException: Attempt to invoke virtual method 'void org.eyeseetea.malariacare.layout.adapters.survey.navigation.NavigationController.first()' on a null object reference
2018-10-26 09:27:24.168 25949-25949/org.eyeseetea.surveillance_cnm_dev W/System.err:     at org.eyeseetea.malariacare.layout.adapters.survey.DynamicTabAdapter.initNavigationController(DynamicTabAdapter.java:205)
2018-10-26 09:27:24.168 25949-25949/org.eyeseetea.surveillance_cnm_dev W/System.err:     at org.eyeseetea.malariacare.layout.adapters.survey.DynamicTabAdapter.<init>(DynamicTabAdapter.java:152)
[...]

PR #2321, opened 2018-10-26
Check if the navigationcontroller is null after show survey and re-lo…

idelcano:
…ad if is null to avoid error after application restart

### :pushpin: References
* **Issue:** connect  #2314
* **Related pull-requests:** 

Important: you need test it before merge development into this branch.

Fix infinite question load (with silent crash in logcat) after crash pressing a survey option.

I added a navigation builder validation before open the survey, if the navigation builder doesn't exist on session the navigation builder must be created, and the navigation builder listener added to show the survey at the end.

Important: you need test it before merge development into this branch.

This commit solves the crash used to test this bug:

https://github.com/EyeSeeTea/pictureapp/pull/2316/commits/9f86a492db09df2f8b0af7ddebc926e6c09c1313
that commit is not merged in this branch to allow replicate the problem.

To force the bug you can move to develompent, revert that commit (only in local), run the app and follow the steps in Use Case 1. 

After that, move to this branch, follow the steps in the use case 1 and check the problem resolved.

- [ ] **Use case 1:**  Open the app and move to settings, move to dashboard, create a survey and press in some crashed option. Click on restart app on the error dialog. The app is restarted in settings, move from Settings activity to dashboard, create a new survey, the load of the survey never ends.

### :art: UI changes?

### DIFF
--- a/app/src/cnm/java/org/eyeseetea/malariacare/data/database/datasources/strategies/AuthenticationLocalDataSourceStrategy.java
+++ b/app/src/cnm/java/org/eyeseetea/malariacare/data/database/datasources/strategies/AuthenticationLocalDataSourceStrategy.java
@@ -1,8 +1,12 @@
 package org.eyeseetea.malariacare.data.database.datasources.strategies;
 
+import android.content.Context;
+
+import org.eyeseetea.malariacare.R;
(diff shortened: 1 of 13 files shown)


xurxodev (approved):
thanks @idelcano I have refactored a little the code to more readable

PR #2329, opened 2018-11-06
Maintenance bug not show downloaded languages

### LINKED ISSUE #2324: Spanish translated in Connect but still not appearing
Spanish has already been translated in POEditor but it doesn't appear in the list and user cannot select it.

manuelplazaspalacio:
### :pushpin: References
* **Issue:** close #2324 
* **Related pull-requests:** 

Solve bug that not shows all downloaded languages.

Getting the downloaded languages in settings activity instead of the fixed languages in strings.xml.

- [ ] **Use case 1:** Login in app with 8001 user, enter in settings and change the language, enter in a survey to check if ti's translated. Not al app strings are translated because now they use the static strings  in the strings.xml file.

### :art: UI changes?


xurxodev (approved):
thanks @manuelplazaspalacio

PR #2344, opened 2018-12-05
update the push period to the minimun

### LINKED ISSUE #2340: Check alarm logic
The client is complaining for several minutes waiting for the alarm to be launched. So he's asking for confirmation on what's the logic under the alarm. 

As far as I know, the alarm only has to do with the logic present in the onCreate() function in the BaseActivity that's calling the AlarmPushReceiver class. 

Looking at there, the push period should depend on the variable stored in the strings file
```
    <string name="PUSH_PERIOD" translatable="false">10</string>
```

But it seems like there's something else interacting with the alarm, as it's rare that a push last for only 10 seconds. 

We would need to properly analyze the push and determine how much time we spend on each task...like:
- xxx secs waiting for alarm
- xxx secs converting the vouchers
- xxx secs checking for the quarantine
- ...

We should end up with something meaningful explaining why the app's not lasting for 10 seconds on each retry

idelcano:
### :pushpin: References
* **Issue:** close #2340 

Check the alarm workflow

I change the push period to 60 secs, the minimun allowed by android for the repeating alarms in api>19

I have checked the alarm workflw:

our alarm (with 10 seconds) runs every minute.
I tested using 120 secs and checked that it runs every 2 minutes (works as expected).

All this periods are not exact:

about <1 sec converting one voucher
about 1 sec for ONE quarantine
about 10 sec for push one voucher
(more info in the issue)

I created a diagram:
untitled diagram

Run the app an check if the push is pushed every 60 secs

### :art: UI changes?

### DIFF
--- a/app/src/main/res/values/donottranslate.xml
+++ b/app/src/main/res/values/donottranslate.xml
@@ -19,7 +19,7 @@
     <string name="pref_cat_server" translatable="false">pref_cat_server</string>
     <item name="language_code" translatable="false" type="string">language_code</item>
 
-    <string name="PUSH_PERIOD" translatable="false">10</string>
+    <string name="PUSH_PERIOD" translatable="false">60</string>
 
     <string name="font_size_system" translatable="false">system</string>
 


xurxodev (approved):
thanks @idelcano

PR #2345, opened 2018-12-06
Maintenance manual push

### LINKED ISSUE #2341: Manual push
Each time the push is happening in background, spinning arrows appear in the action bar, showing the user that something's happening in the background.

Now we need to keep those arrows permanently in the actionbar, and behave as a button, so the user can decide to manually order a push on demand. 

Apart from this, the button must be disabled for 30 seconds after it finishes, to avoid impatient users retrying and retrying, and acting as a DoS attack.

When the button's disabled, it should be greyed out to clearly show it. 

PS: As you probably will notice, if normal push happens on a 10secs periodicity and the button must be disabled during 30secs then that means that the button would be always disabled. Well, that has to do with the issue #2340. The idea is that it's not currently being an alarm each 10secs, but more something like some minutes, so in the end we would like to end up with an alarm every minute, with 30secs of disabled button.

manuelplazaspalacio:
### :pushpin: References
* **Issue:** close #2341  

Can make manual push clicking on the refresh image.

Adding the last push date to the app info to know if can make manual push, and getting it to know if can make manual push. Enabling and disabling refresh button if can or not make manual push.

Change donottranslate string PUSH_PERIOD to 100 Login with 8001 and try demo push and wait to the manual push enables waiting 30 seconds.
Change donottranslate string PUSH_PERIOD to 100 Login with demo and try demo push and wait to the manual push enables waiting 30 seconds.
Try if after doing manual going out from app an entering manual push activation works well. 

### :art: UI changes?


app/src/ereferrals/java/org/eyeseetea/malariacare/domain/usecase/GetCanMakeManualPushUseCase.java
```diff
@@ -0,0 +1,54 @@
+package org.eyeseetea.malariacare.domain.usecase;
+
+import org.eyeseetea.malariacare.domain.boundary.executors.IAsyncExecutor;
+import org.eyeseetea.malariacare.domain.boundary.executors.IMainExecutor;
+import org.eyeseetea.malariacare.domain.boundary.repositories.IAppInfoRepository;
+import org.eyeseetea.malariacare.domain.entity.AppInfo;
+
+import java.util.Date;
+
+public class GetCanMakeManualPushUseCase implements UseCase {
```
xurxodev: As we have said before, a use case should not return primitive values always entities then this use case has no sense, the logic to decide if a manual push is disabled should be where the data is store to decide. In this case, the data is stored in AppInfo then, the entity AppInfo should have a method  CanMakeManualPush and please create tests for this logic

app/src/ereferrals/java/org/eyeseetea/malariacare/domain/usecase/UpdateLastPushDateUseCase.java
```diff
@@ -0,0 +1,35 @@
+package org.eyeseetea.malariacare.domain.usecase;
+
+import org.eyeseetea.malariacare.domain.boundary.executors.IAsyncExecutor;
+import org.eyeseetea.malariacare.domain.boundary.executors.IMainExecutor;
+import org.eyeseetea.malariacare.domain.boundary.repositories.IAppInfoRepository;
+import org.eyeseetea.malariacare.domain.entity.AppInfo;
+
+import java.util.Date;
+
+public class UpdateLastPushDateUseCase implements UseCase {
```
xurxodev: As we have said before, a use case should not return primitive values always entities then this use case has no sense, please use SaveAppInfoUseCase for this

app/src/ereferrals/java/org/eyeseetea/malariacare/factories/SyncFactoryStrategy.java
```diff
@@ -20,4 +23,14 @@ protected IPullController getPullController(Context context) {
     protected IPushController getPushController(Context context) {
         return new WSPushController(context, new SurveyLocalDataSource());
     }
+
+    public UpdateLastPushDateUseCase getUpdateLastPushDateUseCase(Context context) {
```
xurxodev: Remove this methods after remove use cases

xurxodev (changes requested):
@manuelplazaspalacio there are some things to review on this Pull Request:
Functionality:
 - On automatic push, the sync icon does not realize animation and It is not disabled
Code:
- I have added comments in code

xurxodev (approved):
thanks @manuelplazaspalacio

Issue #2339, opened 2018-12-02
Add condition to online/offline status detection


ifoche:
On 3 situations we need to add an extra-check of the online status:
- Just before a push
- When refreshing web pages or entering them for the first time
- When performing a login

This will be done by using a new endpoint (whose details are still not provided).

ifoche:
Ok, we have an answer about the definition of the endpoint, that should allow us to start with the development. Here it goes:

The exact endpoint might change but it will be of the form:
`[WS_url]/available/`

eg. https://apps.psi-mis.org/eRefWSProd/available/

You will be able to use the GET method.

The response to the endpoint will be:
```
{
   "available": true / false,
   "msg": "A plain text message such as: Available / In Maintenance / Unavailable"  
}
```
It is not implemented yet, but that should allow you to make progresses.

xurxodev:
@idelcano @ifoche This issue is not finished by manu, when we have maintenance hours again we should finish it, I assign it to @idelcano

ifoche:
Please find below the answer with the final endpoint implementation. Apparently we will need to add /api/ to the URL
```
Below api endpoint is now available:
https://apps.psi-mis.org/eRefWSProd/api/available
You can also do these:
https://apps.psi-mis.org/eRefWSProd/api/available?detail=Y
https://apps.psi-mis.org/eRefWSProd/api/available?detail=Y&timeout=1000
if 'detail=Y', possible additional information is:
- requestTime, timeoutUsed, wsSettingCacheRefreshed
'wsSettingCacheRefreshed' only shows up if 'wsSetting' cache were refreshed due to timeout or user request, thus, took a bit more time to response on this api endpoint.
```

PR #2349, opened 2018-12-26
Maintenance check api available

manuelplazaspalacio:
Connect #2339


manuelplazaspalacio:
Checking if the app is available on login and on push, we need to check if login and making push is working well with the new check added, creating and modifying test for it. With push there are some test done with this check but it needs more. With login there are no test.
Checking if the api is available needs an use case, repository, datasource; this is done. We need to do test for all of this classes, to check if it's working correctly.

idelcano:
### :pushpin: References
* **Issue:** close #2339
* **Related pull-requests:**

Add condition to online/offline status detection

I continue the @manuelplazaspalacio  work from his comment. 

Fixed some logic errors, fixed tests, and added some push test and create LoginUseCase tests.

Notes: 
- Run test from Staging variant.
- Run app from Debug variant.

To test the app:

The endpoint does not exist in the server at this moment. We should use breakpoints to test the workflow.

In the file eReferralsApiClient, the method getIfIsApiAvailable gets the available status.
I remove the content of that method by:

`
        ApiAvailable apiAvailable = new ApiAvailable(true, "text");

        if(apiAvailable.isAvailable()){
            throw new AvailableApiException();
        }

        return apiAvailable;
`
And you can use a breakpoint to change the apiAvailable.available value

 **Use case 1:** Run tests using Stagging variant
 **Use case 2:** The push of surveys doesn't work when the apiAvailable is false.
 **Use case 3:** The push of surveys work when the apiAvailable is true.
 **Use case 4:** When you enter in a webview with apiavailable is false, the webview is not reloaded.
 **Use case 5:** When you enter in a webview and refresh it, if the api available is true the webview is reloaded
 **Use case 6:** When you enter in a webview and refresh it, if the api available is false the webview show the server api not available error.
 **Use case 7:** First login shows api available error if the api available is false

### :art: UI changes?

app/src/androidTestEreferrals/java/org/eyeseetea/malariacare/CommonTestPreferencesControl.java
```diff
@@ -0,0 +1,127 @@
...
+import org.eyeseetea.malariacare.data.database.utils.PreferencesState;
+import org.eyeseetea.malariacare.data.repositories.ProgramRepository;
+import org.eyeseetea.malariacare.domain.entity.Credentials;
+import org.eyeseetea.malariacare.domain.entity.Program;
+import org.eyeseetea.malariacare.domain.entity.UserAccount;
+import org.junit.After;
+
+import java.io.IOException;
+
+public class CommonTestPreferencesControl {
```
xurxodev: @idelcano in these cases it is preferable to favour the composition over the inheritance.
You can see this article for one more extensive explanation http://xurxodev.com/favoreciendo-la-composicion-sobre-la-herencia-tambien-en-los-test/
please refactor this class to a junit rule

app/src/androidTestEreferrals/java/org/eyeseetea/malariacare/data/LoginUseCaseShould.java
```diff
@@ -0,0 +1,178 @@
...
+
+    private static final String AUTH_OK = "auth_ok.json";
+    private static final String API_AVAILABLE_OK = "api_available_ok.json";
+    private static final String API_AVAILABLE_NO_OK = "api_available_no_ok.json";
+    private CustomMockServer mCustomMockServer;
+    private LoginUseCase mLoginUseCase;
+    private int countSync;
+
+    @Rule
+    public ActivityTestRule<LoginActivity> mActivityRule = new ActivityTestRule<>(
```
xurxodev: To test LoginUseCase is not necessary launch LoginActivity

app/src/androidTestEreferrals/java/org/eyeseetea/malariacare/data/LoginUseCaseShould.java
```diff
@@ -0,0 +1,178 @@
...
+            LoginActivity.class);
+
+    @Before
+    public void cleanUp() throws IOException {
+        mCustomMockServer = new CustomMockServer(new AssetsFileReader());
+        savePreviousPreferences();
```
xurxodev: On this case to use user preferences is not necessary, here the test is more simple if you use a mock using mockito to replace AuthenticationLocalDataSource.

app/src/androidTestEreferrals/java/org/eyeseetea/malariacare/data/sync/exporter/PushUseCaseShould.java
```diff
@@ -47,25 +40,150 @@
...
-
     private Context mContext;
 
+    @Before
+    public void cleanUp() throws IOException {
+        mContext = InstrumentationRegistry.getTargetContext();
+        mCustomMockServer = new CustomMockServer(new AssetsFileReader());
+        savePreviousPreferences();
+
+        SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(
```
xurxodev: I think using mocks, on this tests is not necessary to use real share preferences

app/src/androidTestEreferrals/java/org/eyeseetea/malariacare/data/sync/exporter/WSPushControllerShould.java
```diff
@@ -37,9 +33,11 @@
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 
-public class WSPushControllerShould {
+public class WSPushControllerShould extends CommonTestPreferencesControl {
```
xurxodev: I think using mocks, on this tests is not necessary to use real share preferences

app/src/ereferrals/java/org/eyeseetea/malariacare/fragments/WebViewFragment.java
```diff
@@ -143,31 +145,53 @@ private void loadUrlInWebView() {
...
+                        new GetWebAvailableUseCase.Callback() {
+                            @Override
+                            public void onSuccess() {
+                                executeOnWebAvailable();
+                            }
+
+                            @Override
+                            public void onError() {
+                                if(shouldShowError) {
+                                    showError();
```
xurxodev: In this case, we should show the API Status message to the user

app/src/main/java/org/eyeseetea/malariacare/LoginActivity.java
```diff
@@ -377,6 +377,12 @@ public void onMaxLoginAttemptsReachedError() {
                 Log.d(TAG, "onMaxLoginAttemptsReachedError");
                 mLoginActivityStrategy.disableLogin();
             }
+
+            @Override
+            public void onServerNotAvailable() {
+                Log.d(TAG, "onServerNotAvailable");
+                mLoginActivityStrategy.onServerNotAvailable();
```
xurxodev: In this case, we should show the API Status message to the user

xurxodev (changes requested):
@idelcano I have requested you some changes

app/src/androidTestEreferrals/java/org/eyeseetea/malariacare/rules/PreferencesDependenciesRules.java
```diff
@@ -0,0 +1,148 @@
...
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+import org.mockito.Mock;
+
+import java.io.IOException;
+import java.util.Date;
+
+import static org.mockito.Mockito.when;
+
+public class PreferencesDependenciesRules implements TestRule {
```
xurxodev: @idelcano this class has no sense because has many responsibilities:
1 - Mock server logic, It's good create a rule specific for this as in example on this article http://xurxodev.com/favoreciendo-la-composicion-sobre-la-herencia-tambien-en-los-test/
2 - givenALoginUseCase is used only from LoginUseCaseShould, so I think this method  should be where is called -> LoginUseCaseShould
3 - With others for pushusecase and pushcontroller occur the same

xurxodev (changes requested):
@idelcano there are several things to review:
- Executing androidTest there are test failing
- I have inserted comments in code
- Review use case tests because if It' is not used context because local data sources is mocked then I think that this test can be unit test (no instrumented test) and more speed to execute without emulator

idelcano:
@xurxodev 

After installing the app (using debug variant), log in, create a survey and running the instrumental test(using the staging variant) an exception crash make the test fail sometimes, 
but I can't replicate this problem all the executions...

I modified the BaseActivityShould thread treatment to use the Expresso tools, and I can't replicate the bug anymore. I think it is fixed

The failing test:
update_data_with_new_updated_configuration_version

This test starts with database objects from BaseActivityShould test.

The exception:

```
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String org.eyeseetea.malariacare.domain.entity.Program.getId()' on a null object reference
at org.eyeseetea.malariacare.data.sync.exporter.ConvertToWSVisitor.getConfigFileVersion(ConvertToWSVisitor.java:90)
at org.eyeseetea.malariacare.data.sync.exporter.ConvertToWSVisitor.init(ConvertToWSVisitor.java:71)
at org.eyeseetea.malariacare.data.sync.exporter.ConvertToWSVisitor.<init>(ConvertToWSVisitor.java:60)
at org.eyeseetea.malariacare.factories.SyncFactoryStrategy.getPushController(SyncFactoryStrategy.java:46)
at org.eyeseetea.malariacare.factories.ASyncFactory.getPushUseCase(ASyncFactory.java:34)
at org.eyeseetea.malariacare.services.strategies.PushServiceStrategy.executePush(PushServiceStrategy.java:207)
at org.eyeseetea.malariacare.services.strategies.PushServiceStrategy.onCorrectCredentials(PushServiceStrategy.java:149)
at org.eyeseetea.malariacare.services.strategies.PushServiceStrategy.access$000(PushServiceStrategy.java:35)
at org.eyeseetea.malariacare.services.strategies.PushServiceStrategy$1.onLoginSuccess(PushServiceStrategy.java:76)
at org.eyeseetea.malariacare.domain.usecase.LoginUseCase$4.run(LoginUseCase.java:162)
at android.os.Handler.handleCallback(Handler.java:790)
at android.os.Handler.dispatchMessage(Handler.java:99)
at android.os.Looper.loop(Looper.java:164)
at android.app.ActivityThread.main(ActivityThread.java:6494)
at java.lang.reflect.Method.invoke(Native Method)
at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
```


I fixed another bug after running tests and re-open the app, in that the user was null,
 but I read a write in preferences like other preferences in baseactivityshould test, and it works.

xurxodev (approved):
thanks @idelcano

Issue #2360, opened 2019-01-11
Refactor process to download languages to Clean Architecture


xurxodev:
Refactor according to:
 - UseCase with async executor and main executor
- Create one repository by domain root entity
- Create the repositories with ReadPolicy invoking local and remote data sources when is necessary
- From UseCase inject all dependencies

PR #2361, opened 2019-01-11
Download current language terms first and then the rest in async way

### LINKED ISSUE #2352: On app start download only login activity terms
The App is downloading languages during its boot time. This performance issue consist on downloading only login terms during boot time, enough to show the login screen and then while presenting the login screen download in background the full languages

xurxodev:
### :pushpin: References
* **Issue:** close #2352 
* **Related pull-requests:** 

Optimize the time of splash activity is showing until navigate to the login activity

###   :gear: branches 
**app**:
       Origin: feature-optimize_the_download_of_languages_time_in_splash_activity Target: v1.4_connect

I added a language parameter to DownloadLanguageTranslationUseCase and first download terms of the current language and then download rest languages terms in an async way.

**Note: According to tests on my emulator, the navigation from splash activity to login activity is reduced over 10-15 seg approximately. For poor connections, the improvement can be bigger.**
The architecture corresponds to this process has bad practices, I have created a new issue to refactor in the future. https://github.com/EyeSeeTea/pictureapp/issues/2360

 **Use case 1:** Remove current app, open Logcat and execute. Read traces and confirm that during splash activity visualization only downloads current language and when to navigate to login activity in an async way the rest of languages terms is downloading.

### :art: UI changes?

### DIFF
--- a/app/src/ereferrals/java/org/eyeseetea/malariacare/data/sync/importer/LanguageDownloader.java
+++ b/app/src/ereferrals/java/org/eyeseetea/malariacare/data/sync/importer/LanguageDownloader.java
@@ -2,6 +2,7 @@ package org.eyeseetea.malariacare.data.sync.importer;
 
 
 import android.support.annotation.NonNull;
+import android.util.Log;
 
 import com.raizlabs.android.dbflow.sql.language.SQLite;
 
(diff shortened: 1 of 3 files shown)


ifoche (approved):
@xurxodev I'm having problems with my phone in French. French is a special case because it's present but there is no translated term. I can see a long exception, but the app seems to work and it definitively seems to save time at the beginning, so creating a separate issue to better investigate that problem, but let's consider it as one of the issues we will review after the testing. Thanks for this!

PR #2387, opened 2019-02-01
Fix show multiple soft login dialogs

### LINKED ISSUE #2374: Multiple soft-login dialogs overlapping
Start the app, switch off the screen multiple times and the result is multiple soft-login dialogs overlapping

xurxodev:
### :pushpin: References
* **Issue:** close #2374 
* **Related pull-requests:** 

Avoid multiple soft-login dialogs overlapping

###   :gear: branches 
**app**:
       Origin: bug-avoid_show_multiple_soft_login_dialogs:  Target: v1.4_connect
**bugshaker-android**:
**EyeSeeTea-sdk**:
       Origin: Development
       
- From dashboard activity, I check if exists a previous soft login and on this case, I don`t show a new soft login

**Use Case 1**:  Open app and with a phone without lock pattern turn off and turn on the screen several times, then only one soft login dialog should be visible

### :art: UI changes?

- [] Yeap, here you have some screenshots


ifoche (approved):
thanks @xurxodev

PR #2389, opened 2019-02-01
Disable login button until pin has a value

### LINKED ISSUE #2375: Login button enabled without any character entered in PIN field

xurxodev:
### :pushpin: References
* **Issue:** close #2375 
* **Related pull-requests:** 

Disable login button until pin has value

Manager server available error from soft login

###   :gear: branches 
**app**:
       Origin: bug-disble_login_until_pin_has_value:  Target: v1.4_connect
**bugshaker-android**:
**EyeSeeTea-sdk**:
       Origin: Development
       
- I have added this logic to presenter

**Use Case 1**:  when soft login appears should have login button disabled
**Use Case 2**:  when user type pin login button should be enabled
**Use Case 3**:  when user remove pin login button should be disabled
**Use Case 4**:  when user insert 3 times wrong pin and login is disabled if user type pin on this period login button should maintain disabled until is enabled after 30 seg
**Use Case 5**:  when user insert 3 times wrong pin and login is disabled if user remove pin after 30seg the login button should not be enabled because the pin is empty

### :art: UI changes?

- [] Yeap, here you have some screenshots


ifoche (approved):
thanks @xurxodev

PR #2397, opened 2019-02-06
Verify languages in background if logged user exists

### LINKED ISSUE #2395: If language is already downloaded download terms in background
Splash always download terms no matter if the user already was logged in. It takes too much time to wait for this answer and terms parsing, that in most of the cases doesn't have any update. When a user is already logged in this step should happen in background. Actually the languages translate is composed by 2 steps:
1) Check if languages have been updated
2) Download terms

We only should arrive to step 2 if there's an update, but we can see splash showing "Downloading terms for current language" even when there's no update on terms. 

So in my opinion:
1) First time you open the app it's ok if the app takes its time to download the current language terms before moving to login
2) After that first startup process, terms download should only happen when the language has been updated
3) If what's taking so long is the first call (checking if there are updates in the language by checking the last modification date) then decide to move to background the process when a user's logged in
4) In case it's always downloading the terms for the current language, then move that download to background when there are no updates, and when possible if a user's logged in, take to background even the languages update check

xurxodev:
### :pushpin: References
* **Issue:** close #2395  
* **Related pull-requests:** 

If the logged user exists verify languages only in background

###   :gear: branches 
**app**:
       Origin: feature-verify_languages_always_in_background_if_user_exists v1.4_connect
**bugshaker-android**:
**EyeSeeTea-sdk**:
       Origin: Development
       
- If a logged user does not exist current language is downloaded in sync way and rest of languages in an async way but if user exist all languages are verified in an async way

- [x] **Use case 1:** The first time the current language terms are downloaded in sync way from splash and the rest in an async way
- [x] **Use case 2:** the next times all language terms are downloaded in an async way from splash

### :art: UI changes?

### DIFF
--- a/app/src/ereferrals/java/org/eyeseetea/malariacare/strategies/SplashActivityStrategy.java
+++ b/app/src/ereferrals/java/org/eyeseetea/malariacare/strategies/SplashActivityStrategy.java
@@ -197,7 +197,7 @@ public class SplashActivityStrategy extends ASplashActivityStrategy {
     @Override
     public void downloadLanguagesFromServer() {
         try {
-            if (BuildConfig.downloadLanguagesFromServer && currentUserAccount == null) {
+            if (BuildConfig.downloadLanguagesFromServer) {
                 Log.i(TAG, "Starting to download Languages From Server");
                 CredentialsReader credentialsReader = CredentialsReader.getInstance();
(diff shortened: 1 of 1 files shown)


ifoche (approved):
thanks @xurxodev

Issue #2400, opened 2019-02-14
possibility to disable a field


ifoche:

ifoche:
The idea is to lock fields for manual change.
Either to keep default values or values that have been passed from an "intent in".

This could be configured via a key in the specific data point:
"disabled": "true" (optional and by default value is: "false")

PR #2402, opened 2019-02-15
Maintenance question disabled feature

xurxodev:
### :pushpin: References
* **Issue:** connect #2400 
* **Related pull-requests:** 

The idea is to lock fields for a manual change.
Either to keep default values or values that have been passed from an "intent in".

This could be configured via a key in the specific data point:
"disabled": "true" (optional and by default value is: "false")

###   :gear: branches 
**app**:
       Origin: maintenance-question_disabled_feature: v1.4_connect
**bugshaker-android**:
**EyeSeeTea-sdk**:
       Origin: Development
       
- I have added a new field in question table named disabled
- I have refactored all conversions from API to DB to set this new field
- I have modified DynamicTabAdapter to disable any question with the disabled field equal to true

**Use Case 1**:  Modify any question in server adding new disable property, realize login and create new survey, the question with disable property to true should be disabled

**Use Case 2**:  From Connect2ConnectExampleApp send an intent where the disabled field has value, connect app should open the survey and the disabled field should have the value and be disabled.

### :art: UI changes?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -381,10 +381,10 @@ dependencies {
     compile "com.squareup:javapoet:${libs.javapoetVersion}"
 
     //DBFlow
-    apt "com.github.Raizlabs.DBFlow:dbflow-processor:${libs.dbFlowVersion}"
-    compile "com.github.Raizlabs.DBFlow:dbflow:${libs.dbFlowVersion}"
-    compile "com.github.Raizlabs.DBFlow:dbflow-core:${libs.dbFlowVersion}"
-    compile "com.github.Raizlabs.DBFlow:dbflow-sqlcipher:${libs.dbFlowVersion}"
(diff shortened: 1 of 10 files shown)


ifoche (changes requested):
@xurxodev in [this screenshot](https://raw.githubusercontent.com/EyeSeeTeaBotTest/snapshots/master/android_screenshot114422509275874481.jpg) lastName is disabled, but it doesn't look like. Could you please be sure that disabled fields are greyed to give clearly that information to the user?

xurxodev:
@ifoche Now the background is set to grey when the question is disabled, then when you open a read-only survey where all questions are disabled now all question have the background to grey also.
disabled_field
all_fields_disabled

ifoche (approved):
thanks @xurxodev

PR #2404, opened 2019-02-20
Maintenance refactor from listview to recyclerview

### LINKED ISSUE #2394: Revamp scroll surveys loading performance
We're not currently using the viewholder pattern for dashboard surveys list and that makes the list to be extremely slow when too many surveys are to be shown. We need to speed up that scroll

xurxodev:
### :pushpin: References
* **Issue:** close #2394 
* **Related pull-requests:** 

Improve scroll in survey list using recycler view with view holder

###   :gear: branches 
**app**:
       Origin: maintenance-refactor_from_listview_to_recyclerview Target: v1.4_connect
**bugshaker-android**:
**EyeSeeTea-sdk**:
       Origin: Development
       
I have created a new fragment SurveysFragment with a recycler view
I have implemented a delete survey action by swipe in RecyclerView

Note: DashboardAdapterStrategy is the old adapter login to render surveys item, I have removed the content but I can not remove it to avoid build problems.

**Use Case 1**:  Create many surveys and the scroll should be smooth
**Use Case 2**:  Delete by swipe should work 

### :art: UI changes?

- [ ] Nope, the UI remains as beautiful as it was before!
- [x] Yeap, here you have some screenshots
new_delete_by_swipe


ifoche (approved):
thanks @xurxodev, the performance improvement is huge on a physical device!

PR #2413, opened 2019-03-06
Maintenance change soft login timeout

### LINKED ISSUE #2406: Change soft login auth endpoint timeout
In this order they would prefer:
- Something like what we had when the soft-login was made directly against DHIS2 (my guess is that it's around 2 seconds)
- Otherwise, if what we previously had was higher than 5 seconds, then use 5 seconds instead. 

The idea is that a soft-login waits for the server answer before allowing the user to continue. If a timeout happens then the user is checked against the local copy of the authentication, so they want to limit that to 5 seconds maximum, in order to let them enter the app more quickly

xurxodev:
### :pushpin: References
* **Issue:** close #2406  
* **Related pull-requests:** 

- Assign to soft login against the server a timeout of 5 seconds
- when a timeout is thrown during a soft login then the soft login is verified against local

###   :gear: branches 
**app**:
       Origin: maintenance-change_soft_login_timeout Target: v1.4_connect
**bugshaker-android**:
**EyeSeeTea-sdk**:
       Origin: Development
       
- I have added a new timeout parameter to eReferralsApiclient and when this one is configured to soft login, receive 5 seconds as the timeout.
- I have refactored SoftloginUseCase to verify against last valid credentials when a timeout is thrown from eReferralsApiclient.
- I have added a new test to verify this new logic

1 - Change build variant to ereferralsStaging and execute all unit and instrumentation tests.
2 - Change build variant to ereferralsDebug and remove the app in the device
3 - execute app and verify manually the next use cases:

Change current remoteTimeoutSoftLogin from 5000 milliseconds to lower (getSoftLoginUseCase method in AuthenticationFactoryStrategy).

**Use Case 1**:  start the app realize full login and provoke one action to soft login to appear, for example, close the app and start again. The soft login should work verifying against local. 

### :art: UI changes?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -9,7 +9,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'io.fabric.tools:gradle:1.+'
+        classpath 'io.fabric.tools:gradle:1.27.1'
         classpath 'com.github.dcendents:android-maven-gradle-plugin:1.4'
         classpath 'com.neenbedankt.gradle.plugins:android-apt:1.8'
         classpath 'com.google.gms:google-services:3.0.0'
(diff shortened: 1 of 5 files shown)


ifoche (approved):
thanks @xurxodev

PR #2425, opened 2019-06-12
Avoid saving values when survey is open in review mode:

### LINKED ISSUE #2418: After leaving the form open for some time (~5-10min) "Submit" is impossible
To reproduce, fill a form and leave the app open for 5-10min.
When coming back to it, pressing the submit button is not possible.

### LINKED ISSUE #2420: Jumping between questions is not consistent (next drop downs don't open)
- [ ] Make jumping between questions consistent (for dropdowns and autocomplete open the selection as for date pickers.
- [ ] Add setting to disable jumping

xurxodev:
### :pushpin: References
* **Issue:** close #2418  
* **Related pull-requests:** 

- The real bug is when you open survey in review mode (opened from review fragment) then some values are lost.

###   :gear: branches 
**app**:
       Origin: maintenance/change_keyboard_in_soft_login_to_numeric Target: v1.4_connect
**bugshaker-android**:
**EyeSeeTea-sdk**:
       Origin: Development
       
The lost values are values of child questions because change logic is fired when to assign value in parent . This logic remove child values on save parent process.

I have modified DynamicTabAdapter to avoid saving values when the survey is open in review modeand after loading all values and if is multi-question tab then I deactive review mode

**Use Case 1**:  Create different combinations of surveys and click on cancel button in review screen to go back to the survey in review mode, any value should not be lost.

### :art: UI changes?

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/layout/adapters/survey/DynamicTabAdapter.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/layout/adapters/survey/DynamicTabAdapter.java
@@ -306,11 +306,14 @@ public class DynamicTabAdapter extends BaseAdapter implements ITabAdapter {
         }
     }
 
-
     public void saveTextValue(View view, String newValue, boolean moveToNextQuestion) {
         QuestionDB questionDB = (QuestionDB) view.getTag();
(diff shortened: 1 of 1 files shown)


ifoche (approved):
thanks @xurxodev

PR #2426, opened 2019-06-13
Manage all IOException during push eReferralsAPIClient as NetworkExce…

### LINKED ISSUE #2417: When sync process is interrupted the record turns "red" instead of "orange"
Alpha tester report:

I interrupted the synchronisation of a record voluntarily.
The record turned "red" when I would expect it to turn "orange" and attempt synchronisation again at a later stage.

Note that in some cases (most?) it turns "orange" as expected, but it sometimes does not and the user cannot know whether the record has been properly synchronised or not.

[See here](https://user-images.githubusercontent.com/10115948/58157502-b79dc280-7c78-11e9-9367-40a051c88198.png)

xurxodev:
### :pushpin: References
* **Issue:** close #2417   
* **Related pull-requests:** 

I interrupted the synchronisation of a record voluntarily.
The record turned "red" when I would expect it to turn "orange" and attempt synchronisation again at a later stage.
Note that in some cases (most?) it turns "orange" as expected, but it sometimes does not and the user cannot know whether the record has been properly synchronised or not.

###   :gear: branches 
**app**:
       Origin: maintenance/change_keyboard_in_soft_login_to_numeric Target: v1.4_connect
**bugshaker-android**:
**EyeSeeTea-sdk**:
       Origin: Development
       
The problem is that only SocketTimeoutException and UnknownHostException derivated class from IOException was treated as NetworkException. I have modified the cath to superclass IOException.

Now all IOExceptions are treated as NetworkException in eReferralsAPIClient and when push controller receives a NetworkException then survey status is assigned to Quarantine.

**Use Case 1**:  Create a new survey and during push process change plane mode to on, the survey should appear in brown color (Quarantine) and send it in the next push.

### :art: UI changes?

### DIFF
--- a/app/src/ereferrals/java/org/eyeseetea/malariacare/data/sync/exporter/eReferralsAPIClient.java
+++ b/app/src/ereferrals/java/org/eyeseetea/malariacare/data/sync/exporter/eReferralsAPIClient.java
@@ -115,15 +115,14 @@ public class eReferralsAPIClient {
         } catch (UnrecognizedPropertyException e) {
             ConversionException conversionException = new ConversionException(e);
             wsClientCallBack.onError(conversionException);
-        } catch (SocketTimeoutException | UnknownHostException e) {
+        } catch (IOException e) {
             wsClientCallBack.onError(new NetworkException());
             return;
         } catch (AvailableApiException e) {
             wsClientCallBack.onError(e);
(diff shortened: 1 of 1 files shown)


ifoche (approved):
thanks @xurxodev

PR #2447, opened 2019-10-28
Avoid jump to disabled field

### LINKED ISSUE #2445: In api/available calls, ignore additional keys being returned by the server without failing

xurxodev:
### :pushpin: References
* **Issue:** close #2445                 
* **Related pull-requests:** 

###   :gear: branches 
**app**: 
       Origin: maintenance/jump_error_if_next_is_disabled Target: v1.4_connect
**bugshaker-android**: 
**EyeSeeTea-SDK**: 
       Origin: development   

We should not jump to disabled fields in the survey

- I have added isEnabled getter to IQuestionView interface
- I have implemented isEnabled getter in all custom views
- I have modified jumping logic to avoid jump if the next field is disabled

**UseCase 1**: Login with TZ_TEST_IPC, create a new survey and select Activity = Mjanaja then jumping should work but to found the last disabled field, the jump should be not realized.

### :art: UI changes?


ifoche (approved):
thanks @xurxodev

